PR #4, opened 2011-02-09
setup.py

jemerick:
When installing, the templatetags package was not being installed as it was not included in the packages list.


jsocol:
9d7296a2ba16a633847469def769e779e80df4b7

PR #6, opened 2011-04-12
South migration compatibility

nityaoberoi:
Added south migration files to the project

### DIFF
--- a/waffle/migrations/0001_initial.py
+++ b/waffle/migrations/0001_initial.py
@@ -1,0 +1,122 @@
+# encoding: utf-8
+import datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        
+        # Adding model 'Flag'
+        db.create_table('waffle_flag', (
+            ('id', self.gf('django.db.models.fields.AutoField')(primary_key=True)),
+            ('name', self.gf('django.db.models.fields.CharField')(unique=True, max_length=100)),
+            ('everyone', self.gf('django.db.models.fields.NullBooleanField')(null=True, blank=True)),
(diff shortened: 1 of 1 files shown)


jsocol:
3bf5115a89bbe73b2ab72fcae6c60e20c1bce337

Thanks! I may ask for help adding additional migrations in future versions

nityaoberoi:
Anytime :)

nityaoberoi:
Hi James,

Although you've accepted the pull request above, it doesnt seem to be reflected in the latest version of django-waffle (0.6) on pypi making it difficult for us to adapt to your latest version of the project (since we use south migrations). Any help would be apprecaited. Thanks

jsocol:
Fixed in #8 (I think. We don't use South but it worked on my sqlite test db.)

nityaoberoi:
Not sure why, but it still doesnt reflect on the pypi version:
http://pypi.python.org/pypi/django-waffle/0.7

It seems to be ignoring all new directories - migrations, tests, templates.

jsocol:
Try now with 0.7.1. I switched to using find_packages instead of 
enumerating them.

On 4/20/2011 11:26 AM, nityaoberoi wrote:

> Not sure why, but it still doesnt reflect on the pypi version:
> http://pypi.python.org/pypi/django-waffle/0.7
> 
> It seems to be ignoring all new directories - migrations, tests, templates.
> 
> On Wed, Apr 20, 2011 at 10:58 AM, jsocol<
> reply@reply.github.com>wrote:
> 
> > Fixed in #8 (I think. We don't use South but it worked on my sqlite test
> > db.)
> > 
> > ## 
> > 
> > Reply to this email directly or view it on GitHub:
> > https://github.com/jsocol/django-waffle/pull/6#issuecomment-1033227

## 

James Socol
Support Engineering Manager
Websmith
james@mozilla.com

nityaoberoi:
Looks great! Thanks!

PR #16, opened 2011-07-08
Changed enable/disable switch functionality in djnago admin

etanzapinsky:
Previously when I tried to change the status of a switch by using the dropdown menu in the django admin, the switch's status that was displayed changed but the actual functionality of the change was not reflected in the action that I wanted to perform. For instance, I had a switch that controlled sending a specific type of email. I set it originally that the switch was disabled and made it that the email did not send. When I changed the switch to active using the dropdown, which should have allowed the email to be sent, and tried to send the email, the email also did not send. However, when I enabled the switch and then actually saved it in the django admin, by actively calling the save function, the email was sent out properly. This was also true when I disabled the switch.

I changed the code such that on every time the switch was updated in the django admin via the dropdown it actually calls the save function and doesn't just update the queryset. This got rid of the problem and made it so that the email was sent properly when the switch was updated with the dropdown.

Thanks

### DIFF
--- a/waffle/admin.py
+++ b/waffle/admin.py
@@ -22,12 +22,16 @@ class FlagAdmin(admin.ModelAdmin):
 
 
 def enable_switches(ma, request, qs):
-    qs.update(active=True)
+    for switch in qs:
+        switch.active=True
+        switch.save()
 enable_switches.short_description = 'Enable the selected switches.'
 
 
 def disable_switches(ma, request, qs):
-    qs.update(active=False)
+    for switch in qs:
+        switch.active=False
+        switch.save()
 disable_switches.short_description = 'Disable the selected switches.'
 
 


jsocol:
Calling `.save()` flushes the cache for each switch--which is what needs to happen--but it also means that the admin action will be significantly more expensive when used on large QuerySets. Ideally we could just flush the cache without needing to iterate over the QuerySet.

Of course, how many switches is someone likely to change at the same time? Will need to think a bit about that.

etanzapinsky:
So that's great. I didn't know that flushing the cache was the case since I hava had a simular issue with my own code in other places. 

And regarding your point about how many switches at a time, would there ever be a case when flushing the cache would be slower than re-saving all the switches?

jsocol:
There is no list of stored cache keys to be able to say "flush all switches" right now without iterating over the QuerySet, anyway. I think it makes sense to take this for now and keep the idea of a flush list in mind for the future.

jsocol:
Merged, tagged v0.7.5 and pushed to PyPI, thanks!

PR #24, opened 2011-11-24
Add a way to display a template rather than a simple 404

cyberdelia:
Since a `404` page is not always friendly when turning off temporarily a feature. 
Maybe, we should be able to change status code too, returning a `503` might be a better choice.

### DIFF
--- a/waffle/decorators.py
+++ b/waffle/decorators.py
@@ -1,12 +1,14 @@
 from functools import wraps
 
 from django.http import Http404
+from django.template import RequestContext
+from django.shortcuts import render_to_response
 from django.utils.decorators import available_attrs
 
 from waffle import flag_is_active, switch_is_active
 
 
-def waffle_flag(flag_name):
+def waffle_flag(flag_name, template_name=None):
     def decorator(view):
         @wraps(view, assigned=available_attrs(view))
         def _wrapped_view(request, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


cyberdelia:
Tests are a bit dumb, but I'm can't run them :|

manfre:
503 status would be much better and help prevent a temporarily disabled view from being improperly cached.

jsocol:
503 does make more sense. Seems like that should be specifiable, along with mimetype, too, to be complete. I will take this patch as a starting point and add that.

manfre:
It would be very useful if you provided a means to pass variables through to the template. I've written a similar decorator for flags. See https://gist.github.com/1443360.

jsocol:
At some point I think the level of complexity should indicate something isn't quite right. If you really need that, you have the decorator, but I'm not sure what the use case is in general. I'm curious exactly what you're using Waffle to do, and why you need to give users so much information.

manfre:
The decorator will be applied to several dozen existing views that will allow us to quickly disable portions of the site and will be applied to most new views to allow a more controlled rollout of new apps.

I agree that passing variables thru to the template is probably unnecessarily complex for most usages. The site where I am using the posted decorator suffers from preemptive "usability" push back. Generic informational messages are frequently deemed unfriendly and need to be more specific. The benefit to me of having this functionality in waffle means that I do not need to maintain another slightly modified version of a 3rd party feature. I completely understand if this doesn't make its way in.

jsocol:
I just reopened this to look at it and see if I could do something simple like send a 503 status or mimetype, and I've sort of come to the decision that it isn't worth the complexity, especially since the decorators currently just raise an `Http404`, and returning anything else jumps in complexity (and there is no `HttpResponseServiceUnavailable`).

Ultimately, this is a convenience decorator. If you want more power, you can easily lead a view with:

``` python
def my_waffled_view(request):
    if not waffle.flag_is_active(request):
        return shortcuts.render(request, 'waffled.html', status_code=503)
```

or write a decorator to do the same thing. Once you start getting into all the thing people might want to do here, it gets very complex very quickly. Compared to the relative complexity of doing any one of those things individually, it's just not worth it.

Given how long ago this pull req was opened, I assume you have solutions in place. I'm sorry, it just doesn't make sense to take on this level of complexity within the main project.

Issue #10, opened 2011-04-27
Management commands needed to enable/disable switches


rassie:
I've got an application which has a certain feature which is only useful when a big load of data is loaded into the database. I would like to do something like this:

```
$ ./manage.py switch_off my_feature
$ ./manage.py loaddata fixtures/my_big_pile_of_data.json
$ ./manage.py switch_on my_feature
```

This would obviously require the `switch_on` and `switch_off` commands to be implemented. Any chance this will appear soon?

jsocol:
Interesting idea! It's easy to extend to Samples but difficult to extend to Flags.

rassie:
I actually only thought about Switches. Flags would certainly be possible too with something like `./manage.py flag authenticated on`, but it's way more complicated than Switches and Samples.

jsocol:
It's not the highest priority for me, honestly, but it doesn't seem like a patch would be terribly difficult, especially just for Switches. I'd use syntax like `./manage.py switch my_switch on/off`. Otherwise I might get to it in a 0.9?

rassie:
No urge :-) The patch shouldn't be that difficult indeed, so if I happen to implement it before you do, you'll get a pull request.

mwarkentin:
I've added to what @cyberdelia has done [here](https://github.com/cyberdelia/django-waffle/pull/1)

mwarkentin:
Any thoughts on the pull request? I should have a bit of time to make any updates that you think are needed.

jsocol:
Sorry, I've been swamped the past couple weeks. Let me dig into it today.

mwarkentin:
No worries, whenever you have a chance.

Thanks for the great tool!

jsocol:
Think this was all solved by https://github.com/jsocol/django-waffle/compare/9abbbebb%5E...9c9aa5c

PR #26, opened 2011-11-24
Add a 'switch' and 'sample' command

cyberdelia:
Add a `switch` and `sample` commands that change switch or sample.
This is related to issue #10 but it's missing the `flag` command.

### DIFF
--- a/waffle/management/commands/flag.py
+++ b/waffle/management/commands/flag.py
@@ -7,27 +7,88 @@ from waffle.models import Flag
 
 class Command(BaseCommand):
     option_list = BaseCommand.option_list + (
+        make_option('-l', '--list',
+            action='store_true',
+            dest='list_flag',
+            default=False,
+            help="List existing samples."),
         make_option('--everyone',
-            action='',
-            dest='rollout_activate',
-            help="Activate given target for feature."
+            action='store_true',
+            dest='everyone',
+            help="Activate flag for all users."
         ),
         make_option('--deactivate',
             action='store_false',
-            dest='rollout_activate',
-            help="Deactivate given target for feature."
+            dest='everyone',
+            help="Deactivate flag for all users."
+        ),
+        make_option('--percent', '-p',
+            action='store',
+            type='int',
+            dest='percent',
+            help='Roll out the flag for a certain percentage of users. Takes a number between 0.0 and 100.0'
+        ),
+        make_option('--superusers',
+            action='store_true',
+            dest='superusers',
+            default=False,
+            help='Turn on the flag for Django superusers.'
+        ),
+        make_option('--staff',
+            action='store_true',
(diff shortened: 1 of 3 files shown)


waffle/management/commands/flag.py
```diff
@@ -0,0 +1,33 @@
...
+    def handle(self, flag_name=None, *args, **options):
+        if not flag_name:
+            raise CommandError('You need to specify a flag name.')
+
+        try:
+            flag = Flag.objects.get(name=flag_name)
+            setattr(flag, attribute, value)
+            flag.save()
+        except Flag.DoesNotExist:
+            raise CommandError('This flag doesn\'t exists')
```
mwarkentin: Should be "This flag doesn\'t exist"

You could also do a get_or_create() instead

cyberdelia: I'm don't really want to create a new record from the command line without explicitly asking for it.

mwarkentin: For sure.. I ended up with this in my code:

```
make_option('--create',
            action='store_true',
            dest='create',
            default=False,
            help='If the flag doesn\'t exist, create it.'
        )

if options['create']:
            flag, created = Flag.objects.get_or_create(name=flag_name)
        else:
            try:
                flag = Flag.objects.get(name=flag_name)
            except Flag.DoesNotExist:
                raise CommandError('This flag doesn\'t exist')
```

jsocol: Nit, but when you have apostrophes in a string but no other quotes, it's usually preferable to use double quotes without backslashes. E.g. `"This flag doesn't exist."`. Fixed locally.

jsocol:
I've made a bunch of style changes to make everything PEP8-happy and otherwise consistent (quote, comma styles, etc) with the rest of the code.

I don't love the current `--everyone/--deactivate` switches for Flags, because there's no way to get them back to a Null state, and it's not terribly clear how they're related. I'd like to do something like `--everyone [on/off/something]` instead.

jsocol:
I added [two commits](https://github.com/jsocol/django-waffle/compare/management-commands) and will add documentation tomorrow. If either of you have feedback on the change, let me know!

mwarkentin:
@jsocol Hey James, just wondering if there's any chance this will be making it in at some point? If the documentation is what's holding it up, I could take a swing at it when I have a bit of free time.

jsocol:
Hey, I'm sorry about the delay. I've been trying to catch up on all these projects and PRs. I'll try to get to it today.

mwarkentin:
Awesome, like I said, if there's anything I can help with, let me know.

jsocol:
https://github.com/jsocol/django-waffle/compare/8c5788a...9c9aa5c

Thanks to you both!

PR #28, opened 2011-12-06
WaffleNode.render() raises a KeyError if the context doesn't have a 'request' key. It's helpful for that key to be optional so that we can use waffle samples in email templates.

pdsarin:
...mplates that aren't used in an HTTP context (e.g., email).

### DIFF
--- a/waffle/templatetags/waffle_tags.py
+++ b/waffle/templatetags/waffle_tags.py
@@ -23,7 +23,7 @@ class WaffleNode(template.Node):
             yield node
 
     def render(self, context):
-        if self.condition(context['request'], self.name):
+        if self.condition(context.get('request'), self.name):
             return self.nodelist_true.render(context)
         return self.nodelist_false.render(context)
 


jsocol:
Is there a reason Samples or Switches wouldn't work for this? Flags are explicitly tied to requests because they involve a lot of request/response-specific behavior, like reading and setting cookies, headers, current user, query string, etc.

pdsarin:
I was attempting to use a sample, actually. The code that I patched is in
WaffleNode, which applies to samples too!
On Dec 7, 2011 1:29 AM, "James Socol" <
reply@reply.github.com>
wrote:

jsocol:
Oh, I see, sorry! This is what I get for answering at 1:30am. I will try to get this merged today if I can sort out my network issues.

pdsarin:
Thanks, James!

Phil

On Wed, Dec 7, 2011 at 9:20 AM, James Socol <
reply@reply.github.com

jsocol:
a1ee52939036ee31ff85a68c85a82d61386b95d2

Added a test and merged. Thanks!

PR #31, opened 2012-03-12
Added language support to flags

fgaudin:
To enable releasing new features only for languages which are up to date

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@ dist
 *.egg-info
 *test.db
 .coverage
+.project
+.pydevproject
+.settings
+
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -65,6 +65,11 @@ def flag_is_active(request, flag_name):
     if flag.superusers and user.is_superuser:
         return True
 
+    if flag.languages:
+        languages = flag.languages.split(',')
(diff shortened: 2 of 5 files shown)


jsocol:
I really want to merge this but is `request.LANGUAGE_CODE` standard? At Mozilla we are using `request.locale`. If we're wrong, we're wrong and should switch, but I don't see anything in [the docs](https://docs.djangoproject.com/en/1.4/ref/request-response/#django.http.HttpRequest).

If there's not a standard, we might need a waffle setting to pick the `request` attribute to use here.

fgaudin:
just before this paragraph https://docs.djangoproject.com/en/1.3/topics/i18n/deployment/#how-django-discovers-translations

"Once LocaleMiddleware determines the user's preference, it makes this preference available as request.LANGUAGE_CODE for each HttpRequest. Feel free to read this value in your view code."

jsocol:
Fair enough, we should update our locale middleware to use `LANGUAGE_CODE` instead of `locale`, based on that.

jsocol:
7471bcbcaea2b105e4cf942c0b663f2540511dfc Thanks!

unho:
Please update the docs. There is no reference to this new field on http://waffle.readthedocs.org/en/latest/types.html#flags

PR #35, opened 2012-05-22
Sorting by reverse id for easier administration

henrysxie:
At Yipit we really find this useful as we use waffle switches rather frequently. We care most about the new ones so we just thought this made sense.

I also believe this was the default sort in Django 1.3 but not in 1.4.


jsocol:
Thanks!

PR #36, opened 2012-05-23
Cache non-existant switch

spulec:
Just a small fix and cleanup to the original commit plus some tests.


jsocol:
Thanks for the patch! dcc1781f480e9eb5c9617f4b69f61af5375ca8ad

PR #43, opened 2012-08-07
Fix #12 - Check testing query parameter before cookies.

rlr:
r?

### DIFF
--- a/waffle/tests/test_waffle.py
+++ b/waffle/tests/test_waffle.py
@@ -245,7 +245,7 @@ class WaffleTests(TestCase):
 
         # GET param should override cookie
         request = get(dwft_foo='0')
-        request.COOKIES['dwft_foo'] = True
+        request.COOKIES['dwft_foo'] = 'True'
         assert not waffle.flag_is_active(request, 'foo')
         assert 'foo' in request.waffle_tests
         assert not request.waffle_tests['foo']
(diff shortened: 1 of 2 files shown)


jsocol:
Does this let you reset the testing status (e.g., set it to `1` and then change it to `0` with another request with the GET param)? If so, can you add a test for that?

waffle/tests/test_waffle.py
```diff
@@ -243,7 +243,9 @@ def test_testing_flag(self):
         assert 'foo' in request.waffle_tests
         assert request.waffle_tests['foo']
 
+        # GET param should override cookie
         request = get(dwft_foo='0')
+        request.COOKIES['dwft_foo'] = True
```
jsocol: The cookie should be the string `'True'`, not the actual boolean.

rlr:
Yeah, that is why I added the cookie to the existing test making sure the GET parameter was overriding it. Do you want something in addition to that?

waffle/tests/test_waffle.py
rlr: Ah, yes!

jsocol:
Just a test to demonstrate that you can use the query string to set the testing cookie to one value, then use the query string to set it to another value (doesn't work on master, you have to clear the cookie--I think this patch makes that work).

rlr:
Not sure if new notifications notifies you of updates to the branch now. Anyway, does that test make sense ^

jsocol:
Yay! Whole new (fixed) functionality with the same 2-line code change, thanks! Squashing and merging.

jsocol:
78c34c3789dfdae0272c300da702324659800646

PR #52, opened 2012-09-08
Remove the documentation note about Python/JS function names differing

tobych:
Because they don't, in fact. That's the Django template tags. My bad.


tobych:
Oh dear. It's only commit  fa302ac I'm talking about. I'm not doing this branching right am I. I'll try again.

PR #53, opened 2012-09-08
Remove the doc paragraph about Python/JS function names differing

tobych:
Because they don't. That's the Django template tags. My bad.


jsocol:
e75aba547068d198e181c59804e5d9e12e0e79b5

PR #54, opened 2012-09-08
Change admin's column order so Note is right next to status

tobych:
Moves creation and update timestamps to the right


tobych:
I've just rebased my branch against the upstream master.

jsocol:
ab23d52b911a0f5008a291b7d09052064203be67

PR #55, opened 2012-09-12
Re-add note field for flag admin

kumar303:
the note column was removed from admin after a merge conflict. It was originally added here: https://github.com/jsocol/django-waffle/commit/dd7a5699#L0R20 It was accidentally removed here: https://github.com/jsocol/django-waffle/commit/7471bcbc#L2R20 Note that pull request 54 will be a conflict; it changes the order of the note column but that could be applied by hand.

### DIFF
--- a/waffle/admin.py
+++ b/waffle/admin.py
@@ -23,7 +23,7 @@ class FlagAdmin(admin.ModelAdmin):
     actions = [enable_for_all, disable_for_all]
     date_hierarchy = 'created'
     list_display = ('name', 'everyone', 'percent', 'superusers', 'staff',
-                    'authenticated', 'languages')
+                    'authenticated', 'note', 'languages')
     list_filter = ('everyone', 'superusers', 'staff', 'authenticated')
     raw_id_fields = ('users', 'groups')
     ordering = ('-id',)


tobych:
Pull Request 54 is mine. I'll rebase it against jsocol's master. Right?

tobych:
Ah, no, you shouldn't rebase something you've already pushed to a public repo.
I won't do anything then.
But I could create a new branch and rebase that to the updated master _then_ push that and issue a new pull request...
What's the convention here?

matthewryanscott:
As long as it's a branch that no-one is relying on, even if it's already pushed, it's alright to force push after a rebase.  Here's where I learned about that:  https://github.com/mxcl/homebrew/pull/10030#issuecomment-3853750

Relevant portion:

> No need to submit a new pull request - you can reset the commits and force-push back into the same Github branch, and this pull request will be automagically updated! (Make sure you comment when you do though, Github doesn't notify on new commits.)

tobych:
Thanks Matthew. I reckon I'll do that.

tobych:
Okay, I just rebased my 'columns' branch against the upstream master. So as I write, you can do a fast-forward merge onto master. So no conflicts.

Here's what I did:

```
$ git fetch --all
$ git checkout columns
$ git rebase master
First, rewinding head to replay your work on top of it...
Applying: Removes the doc paragraph about Python/JS function names differing
Applying: Change admin's column order so Note is right next to status
Applying: Removes the doc paragraph about Python/JS function names differing
Using index info to reconstruct a base tree...
Falling back to patching base and 3-way merge...
No changes -- Patch already applied.
Applying: Change admin's column order so Note is right next to status
Using index info to reconstruct a base tree...
Falling back to patching base and 3-way merge...
No changes -- Patch already applied.
$ git push --force origin columns
```

So we're good now. But I'm new to all this so feedback is welcome!

jsocol:
f6d7105045b1225822601cd5787910281fa7582d

PR #69, opened 2013-01-20
Custom user compatibility layer for Django 1.5

thoas:
Hi,

I have added a simple compatibility layer to use the new custom user introduced in Django 1.5.

Quick note: I have also added a TestCase class in **tests/base.py** to provide a default client which is used in your tests, apparently **test_utils.TestCase** redefines **_pre_setup** and does not redeclare a client like [here](https://github.com/django/django/blob/master/django/test/testcases.py#L465).

Let me know if you want further information or elements to include my PR.

### DIFF
--- a/waffle/compat.py
+++ b/waffle/compat.py
@@ -1,0 +1,10 @@
+import django
+
+__all__ = ['User']
+
+# Django 1.5+ compatibility
+if django.VERSION >= (1, 5):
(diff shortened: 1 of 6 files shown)


jsocol:
Just running the tests with 1.5b2 before updating the `.travis.yml` and merging. Thanks!

PR #71, opened 2013-02-22
Add switch_set_active() function

nvie:
This adds a small, but convenient helper function for quickly enabling/disabling switches.

### DIFF
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -128,6 +128,15 @@ def flag_is_active(request, flag_name):
     return False
 
 
+def switch_set_active(name, active):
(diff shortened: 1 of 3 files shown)


nvie:
I've updated this pull request by updating the cached value upon every save. This has the nice benefit of immediate changes, for example when running `python manage.py switch foo on`. Before this, you'd had to wait until the cached `foo=False` value was expired to get the correct value.

jsocol:
I didn't merge this in the last version because I want to:
- Make sure the API is consistent (or as consistent as makes sense) for Flags, Switches, and Samples.
- Use functions like this to eliminate the same code in other places (management commands, admin actions, etc).

PR #76, opened 2013-03-29
Flag user agents

auswm85:
I am working on a project and decided to give waffle a try as my "feature flipper" of choice. I wanted to be able to put certain user agents behind a feature flag so I took a shot at modifying the code to be able to do this. Afterwards I noticed that there was an issue request open for this kind of feature so I thought I would share my solution.

cheers!

### DIFF
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -94,8 +94,8 @@ def flag_is_active(request, flag_name):
 
     if flag.useragents:
         useragents = [u.strip() for u in flag.useragents.split('|')]
-        if (hasattr(request, 'HTTP_USER_AGENT') and 
-                request.HTTP_USER_AGENT in useragents):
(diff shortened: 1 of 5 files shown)


jsocol:
Sorry I'm just looking at this now. Thanks for getting this started (I've wanted it since #3) but there are a couple of minor changes:
- Pipe-delimiting seems more complex than just "one-per-line", especially using a TextField. (Nit, I can fix it.)
- It would be great to make regexes or other fuzzy matches at least possible, if not the default. Being specific is great--sometimes--but UA strings are gnarly and change pretty often.

Do you think you'd have time to take a crack at those? Let me know if not and I'll try to get to it on top of this branch.

jsocol:
I think I'm going to take this feature in a different direction in an upcoming release.

PR #83, opened 2013-07-02
Fix caching/invalidation of 'all' lists

paulcwatts:
The wafflejs view wasn't storing the ALL lists using keyfmt(), so the post_save/post_delete signals that were using keyfmt weren't invalidating the cache.


jsocol:
Thanks!

PR #86, opened 2013-07-25
Fix Django 1.5 User model importing

izquierdo:
I get a circular import problem in my Django 1.5 app caused by waffle getting the user model incorrectly.

From the [docs](https://docs.djangoproject.com/en/1.5/topics/auth/customizing/#django.contrib.auth.get_user_model):

> When you define a foreign key or many-to-many relations to the User model, you should specify the custom model using the AUTH_USER_MODEL setting.

`waffle.compat.User` is used to define a `ManyToManyField`.

### DIFF
--- a/waffle/compat.py
+++ b/waffle/compat.py
@@ -4,7 +4,7 @@ __all__ = ['User']
 
 # Django 1.5+ compatibility
 if django.VERSION >= (1, 5):
-    from django.contrib.auth import get_user_model
-    User = get_user_model()
+    from django.conf import settings
+    User = settings.AUTH_USER_MODEL
 else:
     from django.contrib.auth.models import User
--- a/waffle/compat.py
+++ b/waffle/compat.py
@@ -1,10 +1,10 @@
 import django
 
-__all__ = ['User']
+__all__ = ['AUTH_USER_MODEL']
 
 # Django 1.5+ compatibility
 if django.VERSION >= (1, 5):
     from django.conf import settings
-    User = settings.AUTH_USER_MODEL
+    AUTH_USER_MODEL = settings.AUTH_USER_MODEL
 else:
-    from django.contrib.auth.models import User
+    AUTH_USER_MODEL = 'auth.User'
--- a/waffle/migrations/0001_initial.py
+++ b/waffle/migrations/0001_initial.py
@@ -3,7 +3,7 @@ import datetime
 from south.db import db
 from south.v2 import SchemaMigration
 from django.db import models
-from waffle.compat import User
+from waffle.compat import AUTH_USER_MODEL
 
 class Migration(SchemaMigration):
 
@@ -34,7 +34,7 @@ class Migration(SchemaMigration):
         db.create_table('waffle_flag_users', (
             ('id', models.AutoField(verbose_name='ID', primary_key=True, auto_created=True)),
             ('flag', models.ForeignKey(orm['waffle.flag'], null=False)),
-            ('user', models.ForeignKey(User, null=False))
+            ('user', models.ForeignKey(orm[AUTH_USER_MODEL], null=False))
         ))
         db.create_unique('waffle_flag_users', ['flag_id', 'user_id'])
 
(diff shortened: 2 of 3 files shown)


jsocol:
This makes it possible that `waffle.compat.User` is either a class or a string. The `ManyToMany(AUTH_USER_MODEL)` change certainly makes sense but I don't think this is the way to implement it.

izquierdo:
That's a good point. Since `waffle.compat` only has `waffle.compat.User` which is only used by that `ManyToManyField`, would you agree on removing `waffle.compat` and defining the field as `ManyToManyField(AUTH_USER_MODEL)` instead?

jsocol:
@izquierdo That solution sounds great. One fewer file/modules.

izquierdo:
Sorry, when I proposed that solution I missed the obvious fact that `waffle.compat` is still needed to support django<1.5. So I went for defining `AUTH_USER_MODEL` there instead. Please check this solution and tell me what you think.

jsocol:
@izquierdo And now I need to apologize. The fix for this conflicts with #89, which fixed the south migration for django>=1.5 with a custom user model. I'm not 100% sure how to work around the issue, since South wants a model, not AUTH_USER_MODEL.

izquierdo:
I've pushed my proposed solution that works together with #89.

jsocol:
I got this error with 1.5.5 on this branch.

```
Running migrations for waffle:
 - Migrating forwards to 0008_auto__add_field_flag_languages.
 > waffle:0001_initial
FailedDryRun:  ! Error found during dry run of '0001_initial'! Aborting.
  File "/home/james/.virtualenvs/todaysmeet/lib/python2.7/site-packages/south/migration/migrators.py", line 173, in _run_migration
    migration_function()
  File "/home/james/.virtualenvs/todaysmeet/lib/python2.7/site-packages/south/migration/migrators.py", line 62, in <lambda>
    return (lambda: direction(orm))
  File "/home/james/.virtualenvs/todaysmeet/lib/python2.7/site-packages/waffle/migrations/0001_initial.py", line 37, in forwards
    ('user', models.ForeignKey(orm[AUTH_USER_MODEL], null=False))
  File "/home/james/.virtualenvs/todaysmeet/lib/python2.7/site-packages/south/orm.py", line 172, in __getitem__
    raise KeyError("The model '%s' from the app '%s' is not available in this migration." % (model, app))
KeyError: "The model 'user' from the app 'users' is not available in this migration."
```

kencochrane:
I got the same error, I was able to fix the migration by keeping the User on compat.py with the AUTH_USER_MODEL changes and then using the User  on the django migration. The migrations ran, but that still didn't solve my issue.

My issue is when I try to add a flag via the django admin I get the following error.

> DatabaseError: no such column: waffle_flag_users.myuser_id

Basically the waffle_flag_users table has a user_id field and for some reason the admin is trying to insert the value in myuser_id, most likely because of my custom user model called MyUser. 

At this point I'm not sure what to do, not sure if it is a django bug, or not.

kencochrane:
OK, I got it to work, but I don't know how good of a solution it is, or if it works with django < 1.5. It should work, but I don't have something I can test with. 

Basically I created a custom through table for the Many to Many, and then used this, and it makes it more obvious which table and column to use.  here is what I have

https://github.com/kencochrane/django-waffle/commit/3644c321cb6fe0204a293417eb37f646b0b43baa

[update]
I had to make a few more changes as well, I had to change the admin.py because it doesn't automatically show the users when you use a through table on the many to many. Also, I had to specify the table name for FlagUsers so that I didn't have to create a new migration.  Check my fork for the lastest commits as I work thru everything.

This is getting a little messy so not sure if you want this solution or not. Since I need it to work, I'll just use my fork for now. If you would like me to clean up my commits and submit a PR, let me know.

jsocol:
I feel like the through table has to be overkill for something that must be solved in this setup. I don't have the right environment at hand to test it, but has anyone tried with South 0.8.4? The changelog doesn't go into specifics but 0.8.3, at least, had some issues around custom models.

PR #96, opened 2013-11-03
Python 3 support

coagulant:
Adding python 3.3 support to waffle. 

Due to Fabric still lacking py3 support, I dared to switch Fabric to `invoke` (which is Fabric successor of some kind).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: python
 python:
- - "2.6"
- - "2.7"
- - "3.3"
+ - 2.6
+ - 2.7
+ - 3.3
 env:
  - DJANGO_VERSION=1.4.9
  - DJANGO_VERSION=1.5.5
@@ -12,4 +12,4 @@ script: invoke test
 matrix:
   exclude:
     - python: 3.3
-      env: DJANGO=1.4.9
+      env: DJANGO_VERSION=1.4.9
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,14 @@ language: python
 python:
  - "2.6"
  - "2.7"
+ - "3.3"
 env:
- - DJANGO_VERSION=1.4.8
- - DJANGO_VERSION=1.5.4
+ - DJANGO_VERSION=1.4.9
+ - DJANGO_VERSION=1.5.5
 install:
  - pip install -q Django==${DJANGO_VERSION} -r travis.txt --use-mirrors
-script: fab test
+script: invoke test
+matrix:
+  exclude:
+    - python: 3.3
+      env: DJANGO=1.4.9
--- a/test_app/views.py
+++ b/test_app/views.py
@@ -1,9 +1,8 @@
 from django.http import HttpResponse
-from django.shortcuts import render_to_response
+from django.shortcuts import render_to_response, render
 from django.template import Context, RequestContext
 from django.template.loader import render_to_string
 
-import jingo
 from waffle import flag_is_active
 from waffle.decorators import waffle_flag, waffle_switch
 
@@ -15,7 +14,7 @@ def flag_in_view(request):
 
 
 def flag_in_jingo(request):
-    return jingo.render(request, 'jingo.html')
+    return render(request, 'jingo.html')
 
 
 def flag_in_django(request):
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,46 @@
+[tox]
+envlist = py2.6-django1.4.X, py2.7-django1.4.X, py2.6-django1.5.X, py2.7-django1.5.X, py2.6-django1.6.X,
+          py2.7-django1.6.X, py3.3-django1.5.X, py3.3-django1.6.X
+
+[testenv]
+commands=fab test
+
+[testenv:py2.6-django1.4.X]
+basepython = python2.6
+deps = Django>=1.4,<1.5
+       -rtravis.txt
+
+[testenv:py2.7-django1.4.X]
+basepython = python2.7
+deps = Django>=1.4,<1.5
+       -rtravis.txt
+
+[testenv:py2.6-django1.5.X]
+basepython = python2.6
+deps = Django>=1.5,<1.6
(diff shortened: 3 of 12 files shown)


jsocol:
Wow, this was a lot of lead bullets. Thanks! I'll take and dig in soon.

waffle/__init__.py
```diff
@@ -29,7 +30,7 @@
 def keyfmt(k, v=None):
     if v is None:
         return CACHE_PREFIX + k
-    return CACHE_PREFIX + hashlib.md5(k % v).hexdigest()
+    return CACHE_PREFIX + hashlib.md5(smart_bytes(k % v)).hexdigest()
```
lukaszb, line 33: Why not simply `(k % v).encode()` here?

coagulant, line 33: smart_bytes is what Django uses in similar case: https://github.com/django/django/blob/11cd7388f77aa9d12ab6b57285c3801b237e241b/django/core/cache/utils.py#L14

tox.ini
```diff
@@ -0,0 +1,46 @@
...
+       -rtravis.txt
+
+[testenv:py3.3-django1.5.X]
+basepython = python3.3
+deps = Django>=1.5,<1.6
+       -rtravis.txt
+
+[testenv:py3.3-django1.6.X]
+basepython = python3.3
+deps = https://www.djangoproject.com/download/1.6c1/tarball/
```
ekohl, line 45: 1.6 has been released.

.travis.yml
```diff
@@ -1,10 +1,15 @@
...
- - "2.6"
- - "2.7"
+ - 2.6
+ - 2.7
+ - 3.3
 env:
- - DJANGO_VERSION=1.4.8
- - DJANGO_VERSION=1.5.4
+ - DJANGO_VERSION=1.4.9
+ - DJANGO_VERSION=1.5.5
```
ekohl, line 8: 1.6 has been released as well.

test_app/views.py
```diff
@@ -15,7 +14,7 @@ def flag_in_view(request):
 
 
 def flag_in_jingo(request):
-    return jingo.render(request, 'jingo.html')
+    return render(request, 'jingo.html')
```
ekohl, line 17: Is it still rendered using jingo? Don't you have to use `TEMPLATE_LOADERS` using `@override_settings`?

.travis.yml
lukaszb, line 8: Yeah but let's leave that at another change please.

lukaszb:
Changes look fine to me (I put one minor line note).

@jsocol - can we have this merged? Do you need any more help?

.travis.yml
ekohl, line 8: Because it was added to tox.ini, it may be nice to keep in sync.

lukaszb, line 8: Cannot see tox.ini, not at master at least. And if you meant travis config, it's still another branch (outdated btw).

ekohl, line 8: It's the first commit in the PR: https://github.com/coagulant/django-waffle/commit/d2e3bd879af9dd5b19088b980728d5131824d9d4.

travis.txt
```diff
@@ -2,10 +2,8 @@
...
 mock==1.0b1
-fabric==1.4.3
+invoke==0.5.1
 Jinja2>=2.6
-south==0.7.6
--e git+git://github.com/jbalogh/test-utils.git@3c2214d193d#egg=test-utils
--e git+git://github.com/jbalogh/django-nose.git@e3942c2c9d91267#egg=django-nose
--e git+git://github.com/jbalogh/jingo.git@1dc0e0bf4a0118#egg=jingo
-
+South==dev
```
ekohl, line 7: Already mentioned inline somewhere else, but can this be 0.8.3? I prefer released versions.

waffle/__init__.py
lukaszb, line 33: 1. It actually uses `force_bytes`, which in fact is a lot better (`smart_bytes` does unnecessary check for type of the input parameter, thus is harder to understand why it's used here)
2. We know the input, we know this should not be unicode, that's why we can simply `encode` this. Using `django.utils` is not needed here imho and could fail in future (if at some point Django devs would change it's behaviour, however not probable it sounds currently).

I'm not forcing this, it was simply not obvious for me during review why this particular function was used here. Thought it might not be obvious to someone else too.

test_app/views.py
coagulant, line 17: `test_jingo_tags` method makes sure proper settings are used. https://github.com/coagulant/django-waffle/commit/686df90116cda7ddd87156f417e6af1346059f70#diff-80058049ca008c6d2aa675df888623c3R47

ekohl, line 17: Right. I suppose that at a later point the flag_in_django and flag_in_jingo views can be merged in a later PR, but that's not needed in this PR.

waffle/__init__.py
jsocol, line 33: Calling `encode` here is fine. `django.utils.encoding` has generally been one of the only utils modules with a stable-ish API, but I can't find the relevant list of public modules anymore.

test_app/views.py
jsocol, line 17: Updated by 6563ec530354fd3ecc42d7f67820a2feef4e3b5c.

jsocol:
I've made a couple of changes that will cause collisions but make this a lot simpler:

6563ec530354fd3ecc42d7f67820a2feef4e3b5c - Updates Jingo and a number of other dependencies.
e4dbc21fd9222e543a52e7fda8edf98a34628b77 - Drops Fabric as a requirement for the tests.

Fabric was only setting a couple of environment variables, and it's a very heavy tool for that. That's done by `run.sh` now (I've made similar changes to a number of my other Django-based projects).

I started another branch called [py3k](https://github.com/jsocol/django-waffle/tree/py3k), and you might be able to rebase onto that, or cherry pick the commits here that are still necessary.

coagulant:
Continued in #102

PR #97, opened 2013-11-11
[Feature Proposal] Adding FLAGS, SWITCHES, SAMPLES to window.waffle.

nickpresta:
# Use Case

We have some JavaScript for other services (Qualaroo, Optimizely, etc) that allow us to pass additional metadata. Specifically, we only want to target certain people/groups and to do this, we need to query the flags/switches for a given user.
## Problem

Having the waffle methods isn't enough to be useful as we need to pass raw flag values to the external services so that another person/team in our organization can use the flag names as logic.

We don't want to have to query all the flags/switches/samples for users on our backend and pass it through (possibly in a context processor) as we have quite a few flags/switches and we want to query them as little as possible (for performance reasons).
## Solution

Expose `FLAGS`, `SWITCHES`, and `SAMPLES` objects on `window.waffle`.
## Concerns

I wasn't sure why these raw values weren't exposed -- I thought it may be to discourage people trying to set values (thinking that it would set it on the server) but I'm not sure. If that is the case, I could wrap the objects in a getter and have it return a new object (so that trying to set the values doesn't actually do anything).

### DIFF
--- a/waffle/templates/waffle/waffle.js
+++ b/waffle/templates/waffle/waffle.js
@@ -29,6 +29,9 @@
         if(SAMPLES[sample_name] === undefined) return true;
       {% endif %}
       return !!SAMPLES[sample_name];
-    }
+    },
+    "FLAGS": FLAGS,
+    "SWITCHES": SWITCHES,
+    "SAMPLES": SAMPLES
   };
 })();


coagulant:
Since there is `get_flags` function already (via 62b7d5f8), having FLAGS visible both by getter and window.waffle.FLAGS does not seem DRY to me.

 +1 from me for `window.waffle.FLAGS` for simplicity.

nickpresta:
@coagulant gah! I didn't even notice the `get_flags` function. I agree there should only be one way to access these flags/switches/samples.

I would prefer a `window.waffle.FLAGS/SWITCHES/SAMPLES` for simplicity too, but I don't know if the expectation is that attempting to set `window.waffle.FLAGS` does nothing (other than on that page).

nickpresta:
Also this is basically implemented by: https://github.com/jsocol/django-waffle/issues/81

jsocol:
Thanks @gldnspud, @coagulant and @NickPresta for touching this in various incarnations. Sorry it took me so long to get to it.

PR #98, opened 2013-11-19
Use auto_now and auto_now_add for modified and created

ekohl:
This uses the built in parameters from DateTimeField. See https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.DateField.auto_now for the full documentation.

Note: setting auto_now or auto_now_add to True will cause the field to have editable=False and blank=True set.

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -1,8 +1,3 @@
-try:
-    from django.utils import timezone as datetime
-except ImportError:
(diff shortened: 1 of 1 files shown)


jsocol:
This is intentional. `auto_now` is a huge pain and we skip `auto_add_now` for consistency.

PR #100, opened 2013-12-11
create a wafflejs helper to include without HTTP request

groovecoder:
Found this to be a helpful approach to avoid the extra HTTP request for waffle flags in JavaScript.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -297,8 +297,14 @@ Rollout Mode is enabled **per flag**.
 Waffle in JavaScript
 ====================
 
-Waffle now helps you use flags directly in JavaScript. You need to add
-the Waffle URLs to your URL config::
+Waffle now helps you use flags directly in JavaScript. You can create a global
+``waffle`` object for your JavaScript with either the ``wafflejs`` view, or the
+``wafflejs`` template tag (django) or helper (jinja).
+
+wafflejs view
+-------------
+To use the ``wafflejs`` view, You need to add the Waffle URLs to your URL 
+config::
 
     urlpatterns = patterns('',
         # ...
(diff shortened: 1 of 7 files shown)


groovecoder:
Pull #100 !! w00t!

waffle/helpers.py
```diff
@@ -9,8 +10,14 @@ def flag_helper(context, flag_name):
     return flag_is_active(context['request'], flag_name)
 
 
+@jinja2.contextfunction
+def wafflejs_helper(context):
```
jsocol: This should have the word `inline` in its name somewhere

docs/usage.rst
```diff
@@ -311,6 +320,16 @@ Waffle JavaScript in your templates::
 
     <script src="{% url wafflejs %}"></script>
 
+wafflejs helper
```
jsocol: Waffle has always supported both Jinja and Django templates. How much work would it be to get the helper to work for both?

waffle/views.py
```diff
@@ -11,6 +11,10 @@
 
 @never_cache
 def wafflejs(request):
+    return _generate_waffle_js_resp(request)
+
+
+def _generate_waffle_js_resp(request):
```
jsocol: Is it actually producing an `HttpResponse`, and if so, maybe it's better that the intermediate function return a string and then have the view wrap it in a response object, and the template helper just dump out the string.

docs/usage.rst
groovecoder: This works in Jinja for sure - that's how I'm using it.

jsocol:
#100! :wine_glass: I could've sworn there was a champagne glass emoji

waffle/views.py
groovecoder: Fixed.

waffle/helpers.py
groovecoder: Fixed.

jsocol:
Mind adding some kind of test for both jinja and django templates, just to prove it renders?

groovecoder:
Think I got everything. How's this look for a merge? Would love to stay on the `jsocol` remote for this submodule in kuma-lib. :)

jsocol:
With two new passing tests, I will eagerly hit the blue button :)

groovecoder:
What should I test?

jsocol:
Add a template that uses the helper (within `<script>` tags or not) and just assert that `window.waffle` shows up in the rendered response? See the other template helper tests for a rough outline. Will probably take separate templates for jingo and django.

groovecoder:
There are template tag tests, but no other template helper tests.

jsocol:
[`waffle/tests/test_templates.py`](https://github.com/jsocol/django-waffle/blob/master/waffle/tests/test_templates.py#L21) tests both the Django template tags and the Jingo helpers.

My point here is that this does not support Django templates right now. We can do that with a new [inclusion tag](https://docs.djangoproject.com/en/1.5/howto/custom-template-tags/#inclusion-tags) in `waffle/templatetags/waffle_tags.py`. Tests would demonstrate that it fails. I will merge this as soon as there are two new asserts (preferably new tests) that demonstrate the jingo render path and the django render path actually outputting text.

These can probably be added to the existing test_app templates. For the django inclusion tag, something like `{% waffle_js %}` would be perfect.

groovecoder:
Updated with tests!
tests

jsocol:
Woo, thanks! :beers:

PR #101, opened 2013-12-19
Updated User model reference based on Django recommendations.

vail130:
Changed waffle.compat definition of User object to use
settings.AUTH_USER_MODEL instead of django.contrib.auth.get_user_model().
According to
https://docs.djangoproject.com/en/1.5/topics/auth/customizing/#django.contrib.auth.get_user_model
settings.AUTH_USER_MODEL should be used to define ForeignKey and
ManyToMany relationships instead of the actual User model from
get_user_model().

### DIFF
--- a/waffle/compat.py
+++ b/waffle/compat.py
@@ -1,6 +1,6 @@
 import django
 
-__all__ = ['User']
+__all__ = ['User', 'get_user_model']
 
 # Django 1.5+ compatibility
 if django.VERSION >= (1, 5):
--- a/waffle/compat.py
+++ b/waffle/compat.py
@@ -4,7 +4,7 @@ __all__ = ['User']
 
 # Django 1.5+ compatibility
 if django.VERSION >= (1, 5):
-    from django.contrib.auth import get_user_model
-    User = get_user_model()
+    from django.conf import settings
+    User = settings.AUTH_USER_MODEL
 else:
     from django.contrib.auth.models import User
--- a/waffle/compat.py
+++ b/waffle/compat.py
@@ -5,12 +5,8 @@ __all__ = ['User', 'get_user_model']
 # Django 1.5+ compatibility
 if django.VERSION >= (1, 5):
     from django.conf import settings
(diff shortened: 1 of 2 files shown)


vail130:
As background, the reason I'm proposing this change is because we're using django-waffle (at work), but the way it was referencing the actual User model in its ManyToMany field definition was messing up the results of our code coverage report in our project (we're using Django 1.5.x and implementing our own custom User model). When I changed the field definition to use settings.AUTH_USER_MODEL, our coverage report was back to normal.

As it happens, Django recommends using the settings.AUTH_USER_MODEL instead of the result of django.contrib.auth.get_user_model for ForeignKey and ManyToMany field definitions (https://docs.djangoproject.com/en/1.5/topics/auth/customizing/#django.contrib.auth.get_user_model).

waffle/compat.py
```diff
@@ -1,10 +1,19 @@
...
 
-__all__ = ['User']
+__all__ = ['User', 'get_user_model']
 
 # Django 1.5+ compatibility
 if django.VERSION >= (1, 5):
-    from django.contrib.auth import get_user_model
-    User = get_user_model()
+    from django.conf import settings
+    from django.contrib.auth import get_user_model as get_auth_user_model
```
coagulant: Why not just

```
from django.contrib.auth import get_user_model
```

get_auth_user_model seems redundant?

vail130: Yup. Totally agree.

jsocol:
Thanks! I squashed this and pushed it to master as 605b0b82a76634b03b58ddd151e1bb735fbe167d. If @izquierdo and @kencochrane can verify this fixes #86 for them, I'll wrap up a new release.

kencochrane:
@jsocol this works for me on django 1.6.1 and South 0.8.4, haven't tried with with django 1.5.x where I had the issue in the past.

izquierdo:
I can confirm this works for me in Django 1.5.5.

Thanks!

jsocol:
Fantastic, thanks everyone!

PR #103, opened 2014-01-07
Tox config

coagulant:
Tests against Python 2.6-2.7, and 3.3.
Django 1.4-1.6

If it's ok, I can update [travis.yml to use tox too](http://alexgaynor.net/2014/jan/06/why-travis-ci-is-great-for-the-python-community/).

I've been using tox to develop python 2&3 simultaneous support (#96)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,52 @@
+[tox]
+envlist = py2.6-django1.4.X,
+          py2.7-django1.4.X,
+          py2.6-django1.5.X,
+          py2.7-django1.5.X,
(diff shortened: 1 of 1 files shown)


jsocol:
I'd rather not use tox to run travis, since it seems to prevent travis from parallelizing the jobs.

coagulant:
It doesn't have to prevent parallelization if set up propery: example is https://travis-ci.org/pyca/cryptography

Thnx for merge.

jsocol:
Thanks for the patch!

PR #108, opened 2014-02-03
Implement abstract base class for flags.

mattdeboard:
The original implementation was tightly coupled to Django auth user & group models.
The common fields are now abstracted out into the ABC, allowing users to create their own Flag class
with minimal boilerplate.

This is not a matter of a custom user model, per se. In a project, we are using `User` and `Group` from `django.contrib.auth`, but there are is also a second `User` model that exists on a different db we need to reference. It's complicated, but this will eliminate the need for me (and other projects, I assume) to either copy&paste or perform some hackery so my project can play nice with Waffle.

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -45,6 +45,10 @@ class AbstractFlag(models.Model):
     def __unicode__(self):
         return self.name
 
+    def save(self, *args, **kwargs):
+        self.modified = datetime.now()
+        super(AbstractFlag, self).save(*args, **kwargs)
+
     class Meta:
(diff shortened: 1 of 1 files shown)


mattdeboard:
Ping @jsocol , any input on this here?

mattdeboard:
Actually I guess quite a bit more work would be needed, mostly to abstract the `Flag` class behind a settings variable so the user could specify which model to use, otherwise all the middlewares etc., would just be referring to the original. Gonna close til I find time to dedicate to this.

PR #109, opened 2014-03-13
Possibility to define custom user segments.

sirex:
This commits adds possibility to customize Waffle with possibility to define
your own custom user segments.

For this you need to create a Django form, set it to WAFFLE_CUSTOM_SEGMENT
setting and implement form method called `flag_is_active`. Then in Django
admin, you can set values for fields of this form, these values will be passed
to `flag_is_active` method and this methos will decide if current user falls in
to a custom segment or not.


jsocol:
I want to pursue the ability to do more custom segmenting code, but I don't really believe this is the right approach. It's not per-flag, and it's tacked-on to the existing structure, instead of cleaning it up. An ideal solution would be per-flag, and be designed in such a way it could subsume most, if not all, of the existing per-flag options. Honestly, that may end up being "waffle2".

PR #118, opened 2014-07-08
Add flag/switch/sample template filters

embreinhardt:
Small addition, but makes and/or logic in templates straight forward.

Adds filters:
- `flag` 
  - `[flag_name]|flag:request`
- `sample`
  - `[sample_name]|sample` or `[sample_name]|sample:request`
- `switch`
  - `[switch_name]|switch`, or `[switch_name]|switch:request`

Optional request params on `sample` and `switch` are kept to maintain the pattern used in `flag`, but are unused and can be omitted

### DIFF
--- a/test_app/templates/django/django.html
+++ b/test_app/templates/django/django.html
@@ -40,4 +40,40 @@
   sample_var off
 {% endsample %}
 
+{% if "flag"|flag:request %}
+  flag filter on
+{% else %}
+  flag filter off
+{% endif %}
+
+{% if "switch"|switch %}
+  switch filter on
+{% else %}
+  switch filter off
+{% endif %}
+
+{% if "sample"|sample %}
+  sample filter on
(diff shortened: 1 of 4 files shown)


jsocol:
Ugh, for the lack of something like Jinja2's `@jinja2.contextfunction` decorator.

Couple things:
- Can you run this through `flake8` before we go any further?
- Is the intent here specifically to handle the Django template case of `{% if foo|flag:request and bar|flag:request %}`?
- Let's drop the optional request args to sample and switch, they don't do anything. Unless there's some reason to keep them?
- Needs docs!

stefanfoulis:
An alternative way of doing this would be to put a lazy dictionary into the template context. It would check flags on access on it's keys. Then you could do this:

```
{% if flags.my-flag %}
...
{% endif %}
```

PR #112, opened 2014-04-14
Django 1.7 compatibility

thoas:
- Export signals handling to models.py
- Export all settings to one unique file
- Export keyfmt to utils.py
- We are not importing models anymore in **init**.py which can cause in several cases multiple fails from django.apps Registry
- Refactor unittests using django own DiscoverRunner

Currently all tests pass except SwitchTests.test_no_query which I'm missing something why we should make on query as we don't cache query from Django itself (we only cache proxies from `waffle.__init__.py`).

I'm available if you want more information.

This PR is currently under development and not available to merge btw.

PR #123, opened 2014-08-29
Using a specific cache from settings

### LINKED ISSUE #122: Setting to select a cache
I think it'd be useful to have a setting to determine which cache Waffle should use.

Of course, an optional setting which would default to 'DEFAULT'. I've seen many people using multiple caches, and for Waffle flags I much rather just use local memory than hit my redis cache.

Thoughts?

leonsas:
Waffle will look for the setting WAFFLE_CACHE_NAME and if it exists,
use that cache instead of the default django cache.

This is done in `compat.py`, as Django 1.7 changes the cache getter API a little.

If you think this is a good approach I can add some docs and maybe some tests, except not quite sure how to test this yet.

resolves #122

### DIFF
--- a/waffle/compat.py
+++ b/waffle/compat.py
@@ -1,10 +1,10 @@
 import django
+from django.conf import settings
 
 __all__ = ['User', 'get_user_model']
 
 # Django 1.5+ compatibility
 if django.VERSION >= (1, 5):
-    from django.conf import settings
     from django.contrib.auth import get_user_model
     User = settings.AUTH_USER_MODEL if hasattr(settings, 'AUTH_USER_MODEL') else 'auth.User'
 
@@ -13,3 +13,14 @@ else:
 
     def get_user_model():
         return User
+
+if hasattr(settings, 'WAFFLE_CACHE_NAME'):
+    cache_name = settings.WAFFLE_CACHE_NAME
+    if django.VERSION >= (1, 7):
+        from django.core.cache import caches
+        cache = caches[cache_name]
+    else:
+        from django.core.cache import get_cache
+        cache = get_cache(cache_name)
+else:
+    from django.core.cache import cache
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -3,11 +3,10 @@ import random
 import hashlib
 
 from django.conf import settings
-from django.core.cache import cache
 from django.db.models.signals import post_save, post_delete, m2m_changed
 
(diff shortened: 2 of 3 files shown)


waffle/compat.py
```diff
@@ -13,3 +13,14 @@
 
     def get_user_model():
         return User
+
+if hasattr(settings,'WAFFLE_CACHE_NAME'):
```
jsocol: `settings` isn't imported here if `django.version < (1, 5)`, explaining the 1.4 failures.

Also, pep8: this needs to be `hasattr(settings, 'WAFFLE_CACHE_NAME')` (note the space)

waffle/compat.py
```diff
@@ -13,3 +13,14 @@
...
+    CACHE_NAME = settings.WAFFLE_CACHE_NAME
```
jsocol: variable names like this shouldn't be in all caps

waffle/compat.py
jsocol: and actually this whole thing can be simpler:

``` python
cache_name = getattr(settings, 'WAFFLE_CACHE_NAME', 'default')
if django.VERSION >= (1, 7):
    # ...
```

leonsas: Yeah, just noticed the error, just switched the setting import to the top

jsocol:
Testing `waffle.compat` is tricky as hell, because by definition you'd need to test it against two versions and assert that they both work. I think the fact that all the other tests continue to pass against all versions is sufficient.

As for testing that it changes the cache, I think the only way to do that would be define two caches with different backend classes, because the instances don't seem to have any idea what their name is, then use `@override_settings(...)` to check that the class of `waffle.compat.cache` is different.

Docs would be good, too, yeah :)

leonsas:
I'll try to write a few tests later. As for docs, I'll probably just throw it under Installation.

jsocol:
Eh, actually, skip the docs. I just realized there's no dedicated config chapter, and that's where it belongs, and on me to actually do.

jsocol:
Any chance you could squash these 3 commits into one?

leonsas:
Yeah, I'll try to get that at some point this week.

leonsas:
Done. Are you planning on a PyPI release soon?

jsocol:
Thanks!

I'd like to get #112 merged for a v0.10.1. Updating Github milestones to reflect that now.

PR #125, opened 2014-10-16
Minor docs fix

willkg:
The previous section says that the querystring parameter has to start with `dwft_`,
but the examples don't have this. This fixes the examples.

r?

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -234,8 +234,8 @@ If you turn on the ``WAFFLE_OVERRIDE`` setting, you can guarantee a
 flag will be active for a request by putting it in the query string.
 
 For example, if I use the flag ``example`` in a view that serves the
-URL ``/search``, then I can turn on the flag by adding ``?example=1``
-to the query string, or turn it off by adding ``?example=0``.
+URL ``/search``, then I can turn on the flag by adding ``?dwft_example=1``
+to the query string, or turn it off by adding ``?dwft_example=0``.
 
 By default, ``WAFFLE_OVERRIDE`` is off. It may be useful for testing,
 automated testing in particular.


willkg:
Actually, now I'm confused about this.

One part says that the querystring param has to start with `dwft_`,

https://github.com/jsocol/django-waffle/blame/master/docs/usage.rst#L216

But the part I corrected didn't have the `dwft_`.

Also, I don't see anything like that in the code. Seems like the code looks for the flag name in `request.GET` without prepending anything to it.

https://github.com/jsocol/django-waffle/blob/master/waffle/__init__.py#L58
https://github.com/jsocol/django-waffle/blob/master/waffle/__init__.py#L70

So, what should it be? Should it have the `dwft_` or not? I kind of like that idea since it reduces the likelihood that you have name conflicts between your flag names and querystring params for your application. Plus it makes it clearer you're setting a waffle flag in the querystring. Having said that, I'll sleep fine at night either way.

jsocol:
There are two different situations here. `dwft_%(name)s` is for use with [testing mode](http://waffle.readthedocs.org/en/latest/usage.html#testing-property), while `%(name)s` is for use with the global `WAFFLE_OVERRIDE` setting.

Testing mode for an individual flag is intended for production, to be able to opt specific users in or out of the flag for A/B testing purposes. `WAFFLE_OVERRIDE` is intended for automated tests or continuous integration, allowing tests to set a flag state without needing to create every flag, especially integration tests that may not have direct DB access.

Are the names great? No. The word "testing" gets used _way_ too much in this library but English doesn't have a lot of spare words for this. But they're different things and while the docs could be better—your confusion is understandable—this "correction" is wrong.

willkg:
Aha! That makes more sense. Plus I see the testing code here:

https://github.com/jsocol/django-waffle/blob/master/waffle/__init__.py#L68

I'll re-read the docs and mull over whether I have any improvement suggestions or not.

Thank you!

PR #131, opened 2014-11-05
Drop nose dependency

thoas:
Hi @jsocol 

This PR remove nose dependency everywhere.

PR #132, opened 2014-11-05
Fix custom user usage

thoas:
This PR fixes circular imports by using only the constant `AUTH_USER_MODEL` in `waffle/models.py`.

The south initial migration still use `get_user_model`.


jsocol:
Can you rebase this so we get a build on top of #131?

thoas:
done

jsocol:
:ship:

PR #133, opened 2014-11-05
Export all waffle settings in a separate file

thoas:
This PR introduces a new `waffle/settings.py` file to aggregate all default settings.

### DIFF
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -2,40 +2,28 @@ from decimal import Decimal
 import random
 import hashlib
 
-from django.conf import settings
 from django.db.models.signals import post_save, post_delete, m2m_changed
 
-from waffle.models import Flag, Sample, Switch
-from waffle.compat import cache
+from .models import Flag, Sample, Switch
+from .compat import cache
+from . import settings
+
 
 VERSION = (0, 10)
 __version__ = '.'.join(map(str, VERSION))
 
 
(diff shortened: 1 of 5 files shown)


thoas:
I'm waiting to add the two last PRs:
- Django 1.7 real compatibility
- Export signals to `models.py`

jsocol:
:+1: I'm waiting on Travis but I'm going to tackle these basically in order so far. Let's make sure these are all stable then do the signals change, since that is required for 1.7 but not the other way around, and finally do 1.7 compat?

thoas:
Travis is very slow fiou, I'm waiting for the merge to start working on signals then 1.7 compat as you said :+1:

thoas:
tests passed yeah :tada:

jsocol:
:ship:

jsocol:
Ugh, except I can't automatically merge it. Mind rebasing? (Sorry)

thoas:
done :+1:

jsocol:
Aha! Here we go, failures!

I think the `@override_settings` + `reload(settings)` code path isn't a reliable way to do this. We could try rebasing again on top of #132 but maybe we should just use `mock.patch.object` on `waffle.settings` instead.

Also, would you mind undoing the merge commit and actually rebase it on top of master? It creates a much cleaner and easier to follow history.

thoas:
done @jsocol with mock && rebase, let's wait for the travis result

jsocol:
\o/ Woo!

PR #134, opened 2014-11-05
Export signals methods to models.py

thoas:
This PR moves all signals models from `__init__.py` to `models.py`
and add all models imports in each methods to avoid Django loading errors.


jsocol:
:+1:

PR #138, opened 2014-11-26
Fix broken import in docs about testing

artcz:
Hi, 

I found a small bug in docs, here's a patch that fixes it. 

Cheers,


artcz:
Also added a commit that replaces variables with strings in docs about django templates. In the previous version `flag_name` is a variable name and it's _value_ would be passed as an argument.

jsocol:
Thanks @czartur!

PR #139, opened 2014-11-26
Added integration testing tools

lukaszb:
Please let me know if you're interested in those at all. What's wrong or missing?

### DIFF
--- a/waffle/tests/test_testutils.py
+++ b/waffle/tests/test_testutils.py
@@ -1,0 +1,83 @@
+from django.test import TestCase
+from waffle.models import Switch
+from waffle.testutils import switched
+import waffle
+
+
+class SwitchedTest(TestCase):
+
+    def test_switch_existed_and_was_active(self):
+        Switch.objects.create(name='foo', active=True)
+
+        with switched('foo', active=True):
+            assert waffle.switch_is_active('foo')
+
+        with switched('foo', active=False):
+            assert not waffle.switch_is_active('foo')
+
+        # make sure it didn't change 'active' value
+        assert Switch.objects.get(name='foo').active
+
+    def test_switch_existed_and_was_NOT_active(self):
+        Switch.objects.create(name='foo', active=False)
+
+        with switched('foo', active=True):
+            assert waffle.switch_is_active('foo')
+
+        with switched('foo', active=False):
+            assert not waffle.switch_is_active('foo')
+
+        # make sure it didn't change 'active' value
+        assert not Switch.objects.get(name='foo').active
+
+    def test_new_switch(self):
+        with switched('foo', active=True):
+            assert waffle.switch_is_active('foo')
+
+        with switched('foo', active=False):
+            assert not waffle.switch_is_active('foo')
+
+        # make sure switch is removed (as it was created inside context manager)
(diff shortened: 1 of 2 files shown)


jsocol:
What's the use case here?

lukaszb:
Let's imagine we have a part of the code that depends on multiple switches and want to test multiple scenarios. Usually I'm fine with mocking `waffle.switch_is_active` but when there are multiple switches I would need to _trap_ the mock and put some extra conditions there.

waffle/testutils.py
```diff
@@ -0,0 +1,49 @@
...
+        def test_happy_mode_enabled():
+            ...
+
+    """
+    def __init__(self, name, active):
+        self.name = name
+        self.active = active
+
+    def __call__(self, func):
+        def wrapped(*args, **kwargs):
```
jsocol, line 31: Can you add `@functools.wraps(func)` here?

waffle/testutils.py
```diff
@@ -0,0 +1,49 @@
...
+            with self:
+                return func(*args, **kwargs)
+        return wrapped
+
+    def update(self, active):
+        Switch.objects.filter(id=self.switch.id).update(active=active)
+
+    def __enter__(self):
+        self.switch, self.created = Switch.objects.get_or_create(name=self.name)
```
jsocol, line 40: Huh, I legitimately didn't know you could do that (have expressions on the LHS).

But it makes this line too long.

waffle/tests/test_testutils.py
```diff
@@ -0,0 +1,83 @@
...
+        assert not Switch.objects.get(name='foo').active
+
+    def test_new_switch(self):
+        with switched('foo', active=True):
+            assert waffle.switch_is_active('foo')
+
+        with switched('foo', active=False):
+            assert not waffle.switch_is_active('foo')
+
+        # make sure switch is removed (as it was created inside context manager)
```
jsocol, line 40: This comment (and the one in the next test) is too long and if we replicated the assert before calling either method, I don't think they'd be necessary (i.e. this test assumes that the flag isn't there, so explicitly asserting that pre-condition then re-asserting it is fairly self-documenting).

jsocol:
Ah OK, yeah, I see the value there. I'd like to pursue this but there are a couple of caveats.
- Anything done for one type needs to be done for the others, as long as it makes sense. So `switched` might not be the best name. Maybe following Django's `override_settings` example we should go with `override_{flag,sample,switch}`. It is also grammatically clearer.
- These need docs in the (horribly named) file `docs/testing-waffles.rst`.
- The inline comments I just made.

Totally understand if you're not up for adding similar code for flags and samples, but if not, I'll need to carve out some time to add it myself (or someone else can!) before any of this can make it into a released version.

jsocol:
Thanks for the suggestion and starting code, @lukaszb! Added docs and flag/sample version and tests, and refactored a bit to support those.

lukaszb:
@jsocol hey, thanks, sorry I couldn't help more. Too much going on recently (oh, I'm sure you know what I mean).

jsocol:
I know very much what you mean. It's all good, and it was your idea! :+1:

Issue #77, opened 2013-04-23
wafflejs is (very) query intensive


katzj:
As you add more and more flags, wafflejs becomes increasingly heavyweight due to doing multiple db queries per flag and doing no batching at all.

I've currently got ~ 20 flags/switches and loading wafflejs requires 65 sql queries and 90 cache calls and over a second on my local app.  Making it so that I replicate most of the logic from flag_is_active() and then using prefetch_related to get things into fewer queries gets to 10 queries without any cache calls and cuts the time to 0.3 secs.  But that's not something that's really mainline-able in its current form.  Will try to take a second pass at it unless you have an obvious idea but wanted to put it out there

tobych:
A couple of other feature flag libraries can use Redis as a backend to avoid the DB hits. I imagine Waffle's SQL can be optimized meanwhile, though.

katzj:
waffle already uses django's built-in caching functionality to let you cache.  But doing one query per flag that exists is never going to be optimal, no matter how much caching you put in place :)

jsocol:
I'm definitely interested in a solution to this. I haven't spent a lot of time handling the "lots of flags" case. You could push your not-so-refined solution and we can take it as a starting point.

mwarkentin:
@katzj Would love to see your proof of concept for this.

PR #140, opened 2014-12-08
Expose ability to all flags, switches, samples

jwass:
Add new waffle functions all_flags(), all_switches(), and add_samples().  Move
the code from views.py, add top-level tests.

Basically moves the initial code from `_generate_waffle_js()` to separated functions in `waffle.__init__.py` allowing that ability externally.

### DIFF
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -154,7 +154,7 @@ def all_flags(request):
     if flags is None:
         flags = Flag.objects.values_list('name', flat=True)
         cache.add(keyfmt(settings.FLAGS_ALL_CACHE_KEY), flags)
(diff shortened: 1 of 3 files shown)


jsocol:
Hey @jwass, sorry I haven't gotten around to this yet. I spent most of December surprisingly sick.

I see where you're going here and agree that there's value to an API _like_ this, especially in the context of #77 and #142. But I think in the whole picture, it will need to operate a little differently. One of the goals of an `all_flags`-ish function should be to reduce the number of queries, and I think I just outlined a way to do that in my last comment on 142.

I'm going to close this out but assume that something along these lines will crop up in the solution to 77/142.

jwass:
@jsocol No problem. Thanks. Will definitely be happy to see something like it soon.

PR #141, opened 2015-01-07
feature flipping support for request with session and no user

mwaaas:
I have a project that the user interact with the project via session  I need to add new features to specific user and the only way I can do that is to do flags to sessions.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -8,6 +8,7 @@ usage() {
     echo "  test - run the waffle tests"
     echo "  shell - open the Django shell"
     echo "  schema - create a schema migration for any model changes"
+    echo "  makemigrations - create a schema migration for any model changes in django > 1.7 "
     exit 1
 }
 
--- a/test_app/views.py
+++ b/test_app/views.py
@@ -12,6 +12,11 @@ def flag_in_view(request):
         return HttpResponse('on')
     return HttpResponse('off')
 
+def flag_in_session(request):
+    if flag_is_active(request, 'version1'):
+        return HttpResponse('on')
+    return HttpResponse('off')
+
 
 def flag_in_jingo(request):
     return render(request, 'jingo/jingo.html')
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -30,9 +30,17 @@ def set_flag(request, flag_name, active=True, session_only=False):
     request.waffles[flag_name] = [active, session_only]
 
 
-def flag_is_active(request, flag_name):
+def flag_is_active(request, flag_name, session_key="feature"):
     from .models import cache_flag, Flag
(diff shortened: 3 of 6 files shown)


jsocol:
I can see an argument for adding the `if hasattr(request, 'user')` guard. Waffle intentionally depends only on attributes that `AnonymousUser` should have, but it's possible that if there are never going to be any users, you might not have `AuthenticationMiddleware` installed. But it's incomplete at best, since Waffle still depends on `django.contrib.auth` models.

But the session-key management stuff adds complexity for no real benefit. If you have access to a session, then you already have access to a request—unless this is some <1% case of manipulating the session data through some alternate channel—and to cookies, which are the exact requirements for waffle.

You can use `waffle.set_flag` to change a flag on a specific request.

This is the sort of thing—the <1% cases—I would like to enable people to do in userland code with a more extensible core architecture. But it doesn't belong in the core, which is only going to cover the >90% cases (and doesn't even do all of those, yet).

mwaaas:
if I have request has no user and no cookies it has only session then the waffle won't work.

jsocol:
Django sessions require cookies to work, so the presence of a session implies cookies. Waffle explicitly depends on installing the auth stuff. It might be possible to partially work around that requirement in the short term, but not completely.

Waffle also already supports setting browser-session-only cookies to flag features. You can do something like

``` python
if request.session.get('feature') == 'myflag':
    waffle.set_flag(request, 'myflag', session_only=True)
```

But then you wouldn't really be gaining much over just using `if request.session.get('myflag'):` instead `waffle.flag_is_active(request, 'myflag')`, anyway.

mwaaas:
so if i get a request with session whose key is feature and its value is myflag  I want to implement the new feature but what you are suggesting does not do that

This is what it does: 
if request.session.get('feature') == 'myflag':
    waffle.set_flag(request, 'myflag', session_only=True)

so in the request there is waffles attribute the set_flag sets the waffles flag to the request so the flag will be the next time I receive the same request  ( I want the flag to be active  once a request is receive) then even if I receive the request next time the flag won't be active unless the percent field in flag is set

Then I don't see any complexity in adding this to the code
if hasattr(request, 'session'):
        try:
            flag = Flag.objects.get(name=flag_name)
            if request.session.get(session_key) == flag.name:
                return True
        except Flag.DoesNotExist:
            pass

jsocol:
I'd really recommend not using waffle here, honestly. Just do `if request.session.get('feature') == 'myflag':` instead. Something has to write that into the session, so that's where you'd implement whatever logic to decide who gets the flag.

Implementation issue aside, this just doesn't make sense for waffle. I'm sorry, I don't really have time to spend more time on it.

Issue #137, opened 2014-11-19
Stop using signals to cache objects


jsocol:
Using the signal handlers to cache objects and clear the cache is way too complicated and is even worse when you want to cache misses. It's also very hard to reason about and has been a source of some pain in the past.

The strategy I've been using internally lately is much, much simpler:

``` python
@classmethod
def get(cls, name):
    key = cls._cache_key(name)
    obj = cache.get(key)
    if obj == EMPTY:  # Some sentinel value
        return None  # or raise cls.DoesNotExist
    if obj:
        return obj
    try:
        obj = cls.objects.get(name=name)
        cache.add(key, obj)
    except cls.DoesNotExist:
        cache.add(key, EMPTY)
        return None  # or raise
    return obj

    def save(self, *a, **kw):
        cache.delete(self._cache_key(self.name))
        # ...
```

This works great when, like in waffle, most access is by one particular attribute, in this case the names. This doesn't impact the user API at all, only the internals of some methods.

It's a little more complex with the sub-keys for user and group lists on flags, but I think by being clever with the `_cache_key` and including the `modified` datetime, we can make that invalidation automatic and easier.

jsocol:
I feel bad about this, but here's what I'm about to do.

In `Flag.get()`, what's actually going to get cached is an augmented copy of the instance `.__dict__` with the IDs from the two many-to-many relationships as lists (e.g. `_user_ids` and `_group_ids`). Then the restore-from-cache path is `f = Flag(); f.__dict__.update(cached_dict)`, and two new properties (e.g. `Flag.user_ids`) get added.

Then, for the user/group checks, we only need check that the user ID is in the `flag.user_ids` list, or that the intersection of the user's group IDs and `flag.group_ids` is not empty. I'll do something here to make sure that when we check multiple flags, we don't select the user group IDs every time.

The "all" caches will cache lists of these dicts, instead of whatever they have now.

Finally, signal handlers will be set up on post_save, pre_delete, post_delete, and m2m_changed, for the following:
- _post_save_: update the instance cache and the all cache
- _pre_delete_: update the instance cache with the EMPTY value
- _post_delete_: update the all cache
- _m2m_changed_: update the instance cache and the all cache

The assumption is that changing flags is _very_ rare compared to checking them, so we'd much rather incur these queries predictably on change, instead of on access.

If any of that sends people screaming for the hills... Or if it seems I've forgotten anything here, let me know.

jsocol:
[...]

PR #142, opened 2015-01-07
Support bulk flag checks

mwarkentin:
This is an attempt at fixing the number of queries when checking many flags (see #77 for wafflejs template tags).
### Notes
- I attempted to do this without affecting waffle's API - as you can see, existing tests have not needed to be updated
  - In order to do this, `flag_is_active` has been changed to a wrapper which calls the new `flags_are_active` function with a list consisting of one flag name
  - This allowed me to extract any logic dealing with bulk queries (either DB or cache) out into the `flags_are_active` / `get_flags` functions
- Tests
  - I've only added the one test for the db query check so far
  - I haven't figured out how to test that the number of cache queries is constant yet (likely need to do something with mock)
  - I'm not sure how much coverage the existing tests have, but they're all passing right now
- The naming of some of the functions / variables can probably be improved, this was more of a proof-of-concept

### DIFF
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -59,9 +59,9 @@ def flag_is_active(request, flag_name):
 
 
 def _full_flag_is_active(request, flag):
-    from .compat import cache
+    from .compat import cache, is_string
 
-    if isinstance(flag, basestring):
+    if is_string(flag):
         return settings.FLAG_DEFAULT
 
     if settings.OVERRIDE:
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -49,7 +49,12 @@ def get_flags(flag_names):
 
 def flags_are_active(request, flag_names):
     full_flags, missing_flags = get_flags(flag_names)
-    flags = [(f, _full_flag_is_active(request, f)) for f in full_flags + missing_flags]
+    try:
+        user_groups = request.user.groups.all()
+    except ValueError:
+        user_groups = None
+
+    flags = [(f, _full_flag_is_active(request, f, user_groups)) for f in full_flags + missing_flags]
     return flags
 
 
@@ -59,7 +64,7 @@ def flag_is_active(request, flag_name):
     return flags[0][1]
 
 
-def _full_flag_is_active(request, flag):
+def _full_flag_is_active(request, flag, user_groups):
     from .compat import cache
 
     if settings.OVERRIDE:
@@ -104,7 +109,6 @@ def _full_flag_is_active(request, flag):
         return True
 
     flag_groups = cache.get(keyfmt(settings.FLAG_GROUPS_CACHE_KEY, flag.name))
-    user_groups = user.groups.all()
     for group in flag_groups:
         if group in user_groups:
             return True
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -30,11 +30,30 @@ def set_flag(request, flag_name, active=True, session_only=False):
     request.waffles[flag_name] = [active, session_only]
 
 
+def flags_are_active(request, flag_names):
+    from .compat import cache
+
+    flag_keys = [keyfmt(settings.FLAG_CACHE_KEY, f) for f in flag_names]
+    print flag_keys
+    cached_flags = set(cache.get_many(flag_keys))
+    missing_flags = set(flag_names).difference(cached_flags)
+    print "Cached: {}\nMissing: {}".format(cached_flags, missing_flags)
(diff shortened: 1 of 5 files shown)


jsocol:
I'm kind of loathe to go this direction. It cements and builds on some bad design decisions I made in my younger and more vulnerable days, instead of moving away from those decisions. It also exposes an API I really don't like exposing: `flags_are_active`. (This stuff gets combinatorically hard enough without encouraging people. Checking _n_ flags should be annoying because it's a bad idea, speaking from too much experience.)

I haven't spent a ton of time thinking about this yet, but the sketch in my head is more along the lines of adding an `Flag.is_active(self, request)` method and moving most of the logic from `flag_is_active` into that. We would do something, either at the Manager level or somewhere else, to be more proactive about `prefetch_related` and do a better job caching. (I've also come to hate signals and try to use them as little as possible.)

(We could do similar things with `Switch.is_active(self)` and `Sample.is_active(self)`, making the `*_is_active()` methods much thinner wrappers. Fat models FTW.)

Then, wafflejs would fetch all the flags in one query—caching and invalidating this would be part of "better job caching"—and would only then need to do a slightly cleaner version of

``` python
{f.name: f.is_active(request) for f in flags}
```

Something along the lines of #140 would be helpful to expose that via a nicer API than `Flag.objects.all()` so user code could do the same thing.

Like I said, I haven't spent a ton of time on this, so there may be gotchas or hidden depths I haven't considered yet, but it seems like a move toward a cleaner code base, with way less code in `__init__.py`, and overall better organization/abstraction.

mwarkentin:
Hey James, thanks for taking a look at this.. a few questions / comments:

re: the `flags_are_active` API - this was not really meant as a "check that these 5 flags are all 'on'", rather more like the `all_flags` method (although I return a list instead of a dict) from #140, but with bulk db / cache functionality for better performance when dealing with large number of flags (we currently have ~80 flags). I'm not sure if that helps at all - like I said, my function names were not that thought through. :smile:

To be honest, I haven't fully wrapped my head around how the caching works at a flag / user level, so I'm a little unclear on how far a "better job caching" would get us. If it was improved so that that hit only happened once when the user first logged in (unless something invalidates the list), then I suppose wafflejs would perform a lot better (1 cache hit instead of N).

I definitely like the idea of cleaning up `__init__.py`!

mwarkentin:
@jsocol Here's an example of what `flags_are_active` returns:

``` python
[(<Flag: myflag1>, True), (<Flag: myflag2>, True)]
```

and how wafflejs uses it:

``` python
'flags': [(f[0].name, f[1]) for f in flag_values],
```

jsocol:
My concern about `flags_are_active` is that it's there, and even if it's undocumented, people will use it (and then complain that it isn't documented).

I think a "get all flags" method somewhere is valuable, but we could simplify everything here by creating two paths: one flag or all flags, that handle caching. And for all flags I'd rather it just return the flag objects than also be responsible for checking if they're active—that way it operates more like a building block.

"Better job caching" is related to #137 and just kind of cleaning things up. Amongst the problems with caching right now:
- misses aren't cached at all, because it uses signals
- there are a lot of keys per flag, we could probably drop the users and groups keys if we cached the right thing
- if we can cache the right thing, the all-flags-cache can have the user and group IDs in it 
- that would definitely improve the wafflejs/all-flags case, because we could make 1 cache hit instead of 1+2_n_, and 1 query instead of 3_n_ if it misses.

I totally understand not having your head around what caching does right now: signals are terrible, spooky action at a distance, and really hard to get intuition about. In this dream world, we'd still need at least one, maybe two, handlers on post-save and post-delete, to flush the "all" list and anything that got deleted. (And, since writes to the waffle tables are relatively rare, they could all be write-through, instead of just flushing.)

jsocol:
So my thinking is more along the lines of

``` python
class Flag(...):
    # ...
    def is_active(self, request):

    @classmethod
    def get(cls, name):

    @classmethod
    def get_all(cls):
```

where those methods would return one flag, or a list/queryset of all flags. Then wafflejs would do

``` python
flags = Flag.get_all()
# ...
    'flags': {f.name: f.is_active(request) for f in flags},
```

with analogs for `Switch` and `Sample`.

jsocol:
Anyway, I'm putting some time on my calendar this weekend to dive into this more deeply, now that I think I have a handle on the direction.

mwarkentin:
Ok, that looks pretty sweet. Let me know if there's anything I can help with.. otherwise I'll definitely be keeping an eye on this. :)

PR #145, opened 2015-01-10
Fix #143: Add `waffle_` namespace to management commands

### LINKED ISSUE #143: Namespace management commands
Currently the management commands are `flag`, `sample`, and `switch`. But those are all fairly common words. It'd be more polite to other libraries to do `waffle_flag` or `waffle flag` (the former is easier and perfectly clear).

mwarkentin:
- Update commands
- Update docs


mwarkentin:
This should take care of #143.

jsocol:
Thanks, @mwarkentin!

PR #153, opened 2015-02-25
Support for signals.

pifantastic:
I'd like to be able to fire a Mixpanel event when a user is assigned a flag. To me, signals seemed to be a perfect way to do this. This is a simple implementation, but if feedback is positive I can add more tests as well as support for Samples and Switches.

### DIFF
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -2,6 +2,7 @@ from decimal import Decimal
 import random
 
 from waffle.utils import get_setting, keyfmt
+from waffle.signals import flag_set
 
 
 VERSION = (0, 10, 1)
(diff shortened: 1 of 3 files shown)


jsocol:
Your timing is interesting: I'm actively trying to rip out all the signal-based code I can right now.

I understand your use case here, but why is this better than `set_flag(); tell_mixpanel()` in userspace code?

pifantastic:
Hmm. I guess I was assuming that the calling of `set_flag()` was handled by `django-waffle`, and not intended to be called manually. Where would I do something like that? My own custom middleware? The end goal is to know when exactly a user was assigned a flag, so I could fire the mixpanel event a single time.

jsocol:
`set_flag` is used internally by the dice-roll, yes, but it's exposed explicitly to support less-common flows around activating a flag.

unrelated: what's up with Travis, why hasn't this build finished yet?

pifantastic:
Looks like travis is down for maintenance. 

You're right that I could implement the logic around the flag myself, but it would require me to reproduce a lot of the logic in `flag_is_active`. That's mostly what I was trying to avoid.

pifantastic:
So the requirements of what I'm working on have changed, and I no longer really need signals :)

PR #156, opened 2015-03-18
use django.utils.translation for flag languages

groovecoder:
Flags use `request.LANGUAGE_CODE` to check for languages. That requires `django.middleware.locale.LocaleMiddleware`, which isn't explained in the docs.

This PR changes flag languages logic to use `django.utils.translation` above the `LocaleMiddleware` so that projects using other locale middlewares can still use waffle Flag languages feature.

If this is a WONTFIX, I'll send another PR that adds the `django.middlware.locale.LocaleMiddlware` requirement to the docs and the admin site interface.

### DIFF
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -1,6 +1,8 @@
 from decimal import Decimal
 import random
 
+from django.utils import translation
+
 from waffle.utils import get_setting, keyfmt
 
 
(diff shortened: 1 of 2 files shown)


jsocol:
What happens if `USE_I18N = False`?

groovecoder:
The test fails:

```
======================================================================
FAIL: test_languages (waffle.tests.test_waffle.WaffleTests)
----------------------------------------------------------------------
  File "/Users/lcrouch/code/django-waffle/waffle/tests/test_waffle.py", line 109, in test_languages
    self.assertEqual(b'on', response.content)
AssertionError: 'on' != 'off'
```

So, this feature seems to require either:
- [ ] `django.middleware.locale.LocaleMiddleware` (master branch)

or
- [ ] `USE_I18N = True` (this branch)

Whichever one is best, we should at least document the requirement in the docs.

PR #161, opened 2015-05-18
Allow to overwrite the template tag's name to avoid conflicts.

philippeowagner:
Avoid conflicts with other tags for example the switch/case template tag.

This can be done by defining `WAFFLE_FLAG_TAG_NAME`, `WAFFLE_SWITCH_TAG_NAME` and/or `WAFFLE_SAMPLE_TAG_NAME` in the `settings.py`. This is all optional.

### DIFF
--- a/docs/starting/configuring.rst
+++ b/docs/starting/configuring.rst
@@ -45,3 +45,13 @@ behavior.
 
 ``WAFFLE_CACHE_NAME``
     Which cache to use. Defaults to ``'default'``.
+
+
+``WAFFLE_FLAG_TAG_NAME``
+    Allows to overwrite the template tag name for flag, default is ``flag``.
+
+``WAFFLE_SWITCH_TAG_NAME``
+    Allows to overwrite the template tag name for switch, default is ``switch``.
+
+``WAFFLE_SAMPLE_TAG_NAME``
+    Allows to overwrite the template tag name for sample, default is ``sample``.
--- a/waffle/templatetags/waffle_tags.py
+++ b/waffle/templatetags/waffle_tags.py
@@ -1,6 +1,6 @@
 from django import template
 from django.template.base import VariableDoesNotExist
-
+from django.conf import settings
 from waffle import flag_is_active, sample_is_active, switch_is_active
 from waffle.views import _generate_waffle_js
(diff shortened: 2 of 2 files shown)


philippeowagner:
Hi @jsocol et al. 
Please either merge, specify issues or comment on this PR. Would be great to have this in the main project and have it released soon. We have the conflict mentioned in several project in production. :-/ Thanks.

philippeowagner:
@jsocol – any updates on this?

michaelkuty:
+1

another maintainer which has no time to manage own OS makes me sad @jsocol

philippeowagner:
@michaelkuty me too 
@jsocol please check out https://jazzband.co/

michaelkuty:
@philippeowagner yes, i know, good point @jsocol jazzband is one of recommended way..

jsocol:
Honestly, I think the right solution is to namespace these tags, e.g. `waffle_flag / waffle_switch / waffle_sample`, not to add additional code to make them customizable. (Which isn't just what's here, but would need to include tests.)

philippeowagner:
Renaming the template tags? Seriously? IMHO this is absolutely not the right solution either. This would break backwards compatibility and force everybody using django-waffle to rename each and every `switch`, `sample` and `flag` used. Agreed, the tests are missing. I'm pretty sure that I would have found the time to add some during the last 8 month...

jsocol:
Waffle is still a pre-1.0 project and there's a decent amount of stabilizing to do to get over the 1.0 boundary. Breaking backwards compatibility isn't ideal, but these are badly named and better to fix it pre-1.0 than post.

michaelkuty:
@philippeowagner ok if author has this opinion, we can create friendly fork for us, stable and long term usable with django-comptat, I making integration waffle to other systems and I don't want refactoring all occurrences in next year.. internally I called this attitude: non invasive contribution, it's sad, but sometimes we have no choice

jsocol:
I recognize that you two disagree with this, and that's fine. You are free to fork, but as an over-committed maintainer, I'd encourage you to look at [Gutter](https://github.com/disqus/gutter-django), which seems to have some traction as well.

Issue #152, opened 2015-02-23
Custom User model not working 0.10.1 (Django 1.6)


trikosuave:
We have a custom User model declared exactly as stated in the Waffle docs:

```
class UserProfile(AbstractBaseUser, PermissionsMixin):
```

After installing the app and running migrations (South 1.0), I get the following error when trying to create a Flag object in the admin:

```
OperationalError at /panel/waffle/flag/add/
(1054, "Unknown column 'waffle_flag_users.userprofile_id' in 'field list'")
```

I tried a couple of different things in 0001_initial, compat.py, and models.py, including ideas from the PR #86, with no avail. The only way I was able to make it work was to wipe out the old migrations and create a new initial migration on my system, which then resolved AUTH_USER_MODEL and included the correct models. This solution does not involve any code to submit (except for a new initial migration tailed to my custom user model). 

I am creating this issue to state that django-waffle does not work out of the box with a custom user model and the following versions:

```
django-waffle==0.10.1
Django==1.6.7
South==1.0.2
```

Thank you for any suggestions you may have. I would be glad to test and submit any possible code changes that would alleviate this issue.

jsocol:
This is known and I'm honestly not sure the best way to approach fixing it. There were some mistakes made around the time of Django 1.5 and now we're sort of stuck with these bad migrations. Removing them will break the upgrade path and not removing them breaks the new install path.

I _think_ the most reasonable way to handle this is just to say "screw it" and break the upgrade path from <0.10 to 0.11. That probably means removing the south migrations, regenerating them as a new 0001 (or 000whatever's next) and telling folks to delete the ghost migrations. The Django 1.7 migrations don't have this problem, since they don't have the history.

To do that, I need to double-check the current migrations, when they were added, etc, and figure how exactly what versions will break.

trikosuave:
Thanks! It would be amazing if the migration was fixed for 0.11. I needed that and CBV Mixins which is already a milestone for 0.11. 
I changed another piece in my fork which I would love to get your opinion. In the FlagAdmin I changed:

```
-    raw_id_fields = ('users', 'groups')
+    filter_horizontal = ('users', 'groups')
```

I found it very hard to navigate and keep track of users added, especially when it lists the IDs. I can make another branch and do a Pull Request if you wish, unless you just want to make the change in a current branch of yours, if you like the idea.

Thanks!

mauricioabreu:
Do you need help fixing this?
I am reading this issue because I am having this same error.

[...]

PR #165, opened 2015-05-25
Reseted South migrations

mauricioabreu:
Since any South migrations are not working
this patch fixes Django 1.6 compatibility (even breaking upgrade path).


mauricioabreu:
Related to #152

PR #167, opened 2015-06-04
Add support for django-jinja.

rlr:
This moves the Jinja2 helpers to templatetags/jinja_waffle_tags.py
so that the django infra always imports the file and they get loaded.
It tries using both jingo and django-jinja.

This also adds testing with Django 1.8 to the tox settings. Django 1.8
is only tested with django-jinja.

r?


rlr:
hmm, I need to tweak travis.yml to run django 1.8... On it.

test_settings.py
```diff
@@ -73,3 +57,68 @@
...
+    INSTALLED_APPS += ('django_jinja', )
+
+    TEMPLATES = [
+        # Configure the Jinja2 template engine
+        {
+            'BACKEND': 'django_jinja.backend.Jinja2',
+            'DIRS': [],
+            'APP_DIRS': True,
+            'OPTIONS': {
+                "match_regex": r"jingo.*",
```
willkg, line 94: Nit: This uses " but should use '.

test_settings.py
```diff
@@ -73,3 +57,68 @@
...
+                'match_extension': '',
+                'context_processors': [
+                    'django.template.context_processors.debug',
+                    'django.template.context_processors.request',
+                    'django.contrib.auth.context_processors.auth',
+                    'django.contrib.messages.context_processors.messages',
```
willkg, line 100: I don't think we need the `debug` and `messages` context processors. I'm not sure it makes much of a difference, but I think I would leave them out to match the jingo configuration.

test_settings.py
```diff
@@ -73,3 +57,68 @@
...
+    )
+
+else:
+    # New style TEMPLATES setting.
+    # Use django-jinja
+
+    INSTALLED_APPS += ('django_jinja', )
+
+    TEMPLATES = [
+        # Configure the Jinja2 template engine
```
willkg, line 88: You might want to make this more specific. Something like:

```
Configure the Jinja2 template engine so that it looks at the templates we have in
`templates/jingo/`.
```

waffle/templatetags/jinja_waffle_tags.py
```diff
@@ -0,0 +1,39 @@
...
+    pass
+
+
+try:
+    from django_jinja.library import global_function
+
+    global_function('waffle', helpers)
+except ImportError:
+    # django-jinja isn't being used. Move on.
+    pass
```
willkg, line 39: I can't think of a situation where both jingo and django-jinja would be available, but I think I'd change this to cascade. Something like this:

```
try:
    from django_jinja.library import global_function
    global_function('waffle', helpers)
except ImportError:
    try:
        import jingo
        jingo.env.globals['waffle'] = helpers
    except ImportError:
        pass
```

@jsocol Does that sound ok?

waffle/tests/test_templates.py
```diff
@@ -48,19 +48,7 @@ def test_no_request_context(self):
...
-        'django.template.loaders.app_directories.Loader',
-    ))
-    @mock.patch.object(template.loader, 'template_source_loaders', None)
-    def test_jingo_tags(self):
-        """
-            We're manually changing default TEMPLATE_LOADERS to enable jingo
-
-            template_source_loaders needs to be patched to None, otherwise
-            TEMPLATE_LOADERS won't be overriden.
-        """
```
willkg, line 63: I don't think I understand this code. It looks like in the old and new versions `TEMPLATE_LOADERS` is set in `test_settings.py`. So I don't think this is actually overriding any settings.

Does that suggest this test isn't valid?

waffle/templatetags/jinja_waffle_tags.py
rlr, line 39: In the case here (running tests in travis), both jingo and django-jinja will be in the virtualenv. Unless I change things so they aren't. And even for sumo, we'd have to make sure the virtualenv gets destroyed once we change off of jingo. I don't think our deploys do that every time.

waffle/tests/test_templates.py
rlr, line 63: I deleted this because it didnt make sense to me either. I was thinking this was there before test_settings looked like what they look like now.

rlr, line 63: The test did fail with django 1.8 until I added the django-jinja bits though. So I am pretty sure the test is good.

waffle/templatetags/jinja_waffle_tags.py
willkg, line 39: Mmm... We should fix the test harness so we have separate "jingo" and "django-jinja" environments with different things installed. Otherwise we're mixing things and it's not clear what confidence of correctness our tests can give us.

willkg, line 39: The SUMO problem should get solved in SUMO-land. There's a bug for this problem.

willkg:
I think this looks ok. I had some minor nits but they're probably not that important.

The one thing I was thinking about was whether truncating `waffle/helpers.py` and then moving that code to `waffle/templatetags/jinja_template_tags.py` was a good idea or whether it'd be better to have the jingo specific code in `waffle/helpers.py` and not mix it with the django-jinja stuff in `waffle/templatetags/jinja_template_tags.py`.

Maybe that's academic, though.

@jsocol If there's anything else I can do here, let me know. I don't want to rush you, but the sooner this lands, the better. I'm totally game for trading favors (e.g. you spend some time on this and I spend some time helping you with something else) if that helps.

willkg:
This has been around for a month and a half now. We need to move forward, so I'm going to fork django-waffle for now and we'll use my fork.

If there's anything I can do to help get this landed, let me know.

jsocol:
I added @willkg to this repo—we talked about it offline a little while ago and I think it slipped off both our plates. I'm actively looking for more help maintaining this and a few other repos, too.

That said, the next release has got to be 0.10.2, then I'll get to #137 in 0.10.3, I guess, before wrapping up the work in 0.11.

willkg:
I'm a huge fan of what @jsocol just did with #180. Given that solves the problems we were solving here, I'm closing this one out.

PR #169, opened 2015-07-20
Fixed South initial migration to support custom user models.

### LINKED ISSUE #152: Custom User model not working 0.10.1 (Django 1.6)

combitel:
This patch will fix #152. Hope, this will help somebody.

### DIFF
--- a/waffle/south_migrations/0001_initial.py
+++ b/waffle/south_migrations/0001_initial.py
@@ -16,6 +16,10 @@ else:
     def get_user_model():
         return User
 
+User = get_user_model()
+user_orm_label = '%s.%s' % (User._meta.app_label, User._meta.object_name)
+user_model_label = '%s.%s' % (User._meta.app_label, User._meta.module_name)
+
 
 class Migration(SchemaMigration):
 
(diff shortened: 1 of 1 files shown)


waffle/south_migrations/0001_initial.py
```diff
@@ -46,9 +50,9 @@ def forwards(self, orm):
         db.create_table('waffle_flag_users', (
             ('id', models.AutoField(verbose_name='ID', primary_key=True, auto_created=True)),
             ('flag', models.ForeignKey(orm['waffle.flag'], null=False)),
-            ('user', models.ForeignKey(get_user_model(), null=False))
+            (User._meta.model_name, models.ForeignKey(orm[user_orm_label], null=False))
```
mknecht, line 53: @combitel Why did you make the name of the `user` field dynamic? Isn't this an arbitrary, waffle-internal name that may as well stay hardcoded? If yes, then the 1.4 test failure will be resolved and this PR ready to go. Which would be great, since we have the issue discussed in #152 and fixed here, too. Thanks for the PR! :+1:

mauricioabreu:
This patch is updating the initial migration, right? It is ok?

jsocol:
I didn't do it in purpose, but since 1.4 ESR is no longer supported, and South doesn't support any supported versions anymore, I'm going to close this and replace it with a "drop all the South migrations" issue.

PR #171, opened 2015-08-06
Added support for using override_* decorators on test classes

mark-adams:
Currently, `override_flag`, `override_switch`, and `override_sample` only work as decorators on methods.

This PR makes it so:

``` python
@override_flag('my-flag', active=True)
class MyTestCase(unittest.TestCase)
   def first_test(self):
       ...
   def second_test(self):
       ...
```

is equivalent to:

``` python
class MyTestCase(unittest.TestCase)
   @override_flag('my-flag', active=True)
   def first_test(self):
       ...
   @override_flag('my-flag', active=True)
   def second_test(self):
       ...
```

### DIFF
--- a/waffle/compat.py
+++ b/waffle/compat.py
@@ -1,7 +1,10 @@
 import django
+import sys
+import types
 from django.conf import settings
 
 __all__ = ['cache']
+PY3 = sys.version_info[0] == 3
 
 if hasattr(settings, 'WAFFLE_CACHE_NAME'):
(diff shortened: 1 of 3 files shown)


jsocol:
Thanks!

mark-adams:
Thanks!! When do you think we'll see these in a release?

Last time I checked, the testutils.override_\* decorators haven't yet been included in a release so most people can't take advantage of them :disappointed:

PR #175, opened 2015-09-02
Add SessionKV criteria to percentage based Flag

kew-l:


willkg:
I don't understand what this is about. Is this related to an existing feature request in the issue tracker? If not, can you explain what problem you're trying to solve and why you're doing it this way?

jsocol:
As always, [the roadmap](http://waffle.readthedocs.org/en/latest/about/roadmap.html) is subject to change if there's a really compelling +90% feature, but I'd also like to hear more about the motivation here. This behaves differently than other flag features, and has a number of implementation issues (frex, it could be done without adding new database tables at all).

kew-l:
I need the percentage based Flag to only activate if additional select session key/value pairs match as well.

jsocol:
This is just too far down the wrong road. Adding abstract base classes will allow this kind of customization in user-space, and that's the route forward from here.

PR #178, opened 2015-10-02
Fixed strings in migration

hwkns:
Right now, if you're using Python3 and Django 1.7+, if you install `django-waffle` and run `manage.py makemigrations`, it will create a new migration because the `help_text` on all the fields contains text, not bytes as implied in the initial migration.

This PR changes the initial migration to use unicode literal strings (by removing the `b''` prefix), and updates `models.py` to use unicode strings too (via the `unicode_literals` import), so no extra migrations will be necessary in py2 or py3 environments.


fpruitt:
LGTM

hwkns:
@jsocol any thoughts on this?  Should be a painless change.

jsocol:
Shouldn't we also add `unicode_literals` to `waffle/models.py`? This fix makes sense but it does move the models out of sync with the migrations, on py2.

hwkns:
Ah yes, good catch -- I forgot that Django migrations automatically use `unicode_literals`.  I've added that everywhere, and I also added Django's `python_2_unicode_compatible` decorator to all models.

jsocol:
Thanks!

Issue #179, opened 2015-10-19
update jingo


jsocol:
In django 1.8, `get_standard_processors` moved, and a few other things. as of https://github.com/jbalogh/jingo/pull/74, jingo supports these changes, but not in master yet. to fix [django1.8 tests](https://travis-ci.org/jsocol/django-waffle/jobs/60748561) and merge #159, we'll need to pick up the updated jingo

willkg:
In Mozilla-land, all the projects I'm aware of are dropping Jingo and friends for django-jinja and other apps that are more actively maintained (plus not maintained by us) and are closer to vanilla Django.

I mention that because you seem to be doing all the work on Jingo lately, but if downloads on PyPI are much to go on, then Jingo probably has a pretty small user base. Might be worth ending that project as well as eventually dropping support in django-waffle for it.

jsocol:
Well, it's still supported by waffle for today. If no one is using it there, I might start cowboying stuff in. We should absolutely do what we need to do to support django-jinja, if that's the way the wind is blowing, but I'm not dropping jingo just yet.

jsocol:
Here's the route I'm trying right now: move from mutating the `jingo.env` directly to adding a legit Jinja extension that injects the globals. That should be compatible with any jinja2 template engine.

willkg:
@rlr did a PR for fixing things to support django-jinja #167. It sounds like you want to go a different route (totally fine). I'm game for testing your changes with django-jinja if that helps.

jsocol:
I feel like the jinja2 extension is the "right" way to do this. And it gives us broader support with less hacky stuff, especially [this whole set of comments](https://github.com/jsocol/django-waffle/pull/167/files#diff-fc8be72bf42266bd813ee2fdeec40a0bR39).

But, thanks for reminding me, this largely solves the "ugh testing" problem I just ran into.

jsocol:
Closing this for #180 as well

PR #180, opened 2015-10-20
Django 1.8

jsocol:
@willkg This addresses #179, #167, and probably a few other things.


jsocol:
Ugh, those test failures. I do not entirely understand those. Maybe @rlr recognizes those?

jsocol:
OK, on 1.7, for some reason it's trying to parse the jinja template with django, leading to [this](https://travis-ci.org/jsocol/django-waffle/jobs/86409688#L125):

```
ERROR: test_jinja_tags (waffle.tests.test_templates.WaffleTemplateTests)
----------------------------------------------------------------------
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/django/template/smartif.py", line 173, in translate_token
    op = OPERATORS[token]
KeyError: "waffle.flag('flag')"
...
django.template.base.TemplateSyntaxError: Could not parse the remainder: '('flag')' from 'waffle.flag('flag')'
```

And for some reason the `RequestContext` doesn't have `request` in it:

```
ERROR: test_django_tags (waffle.tests.test_templates.WaffleTemplateTests)
----------------------------------------------------------------------
  File "/home/travis/build/jsocol/django-waffle/waffle/tests/test_templates.py", line 29, in test_django_tags
    response = process_request(request, views.flag_in_django)
...
  File "/home/travis/build/jsocol/django-waffle/waffle/templatetags/waffle_tags.py", line 83, in render
    return _generate_waffle_js(context['request'])
  File "/home/travis/virtualenv/python3.4.2/lib/python3.4/site-packages/django/template/context.py", line 70, in __getitem__
    raise KeyError(key)
KeyError: 'request'
```

jsocol:
Aha, `(VERSION <= (1, 7)) != (VERSION < (1, 8))`

jsocol:
OK, @willkg, @rlr, one of you want to double-check my work here? I'll get something added to the docs asap.

jsocol:
This is a breaking change which would kind of imply we should renumber some milestones, of course. Have to look at that.

waffle/jinja.py
```diff
@@ -15,9 +15,11 @@ def inline_wafflejs_helper(context):
     return _generate_waffle_js(context['request'])
 
 
-jingo.env.globals['waffle'] = {
-    'flag': flag_helper,
-    'switch': switch_is_active,
-    'sample': sample_is_active,
-    'wafflejs': inline_wafflejs_helper
-}
+class WaffleExtension(Extension):
```
rlr, line 18: :+1: I like this approach. Much cleaner than what I came up with. I guess it would need to be tested with django-jinja but I don't see why it wouldn't work.

jsocol, line 18: I stole most of your code from #167, so the 1.8 tests actually do run with django-jinja! :)

rlr, line 18: gah! I see that now. :cake:

jsocol, line 18: thanks for that, btw, it was very helpful!

willkg:
I like this! +1

jsocol:
I'm going to merge this and then spend some time later updating the roadmap and milestones to have new numbers. This basically just becomes v0.11 by itself, or with anything easy to tack on top of it, and everything else moves down one.

a-very-merry-unbirthday-15

PR #183, opened 2015-10-27
Fix typo in docs

alazaro:
From: https://github.com/jsocol/django-waffle/blob/master/waffle/defaults.py#L5
`TEST_COOKIE = 'dwft_%s'`

### DIFF
--- a/docs/starting/configuring.rst
+++ b/docs/starting/configuring.rst
@@ -9,7 +9,7 @@ behavior.
 
 ``WAFFLE_COOKIE``
     The format for the cookies Waffle sets. Must contain ``%s``.
-    Defaults to ``dwf_%s``.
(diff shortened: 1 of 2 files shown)


docs/starting/configuring.rst
```diff
@@ -9,7 +9,7 @@ behavior.
 
 ``WAFFLE_COOKIE``
     The format for the cookies Waffle sets. Must contain ``%s``.
-    Defaults to ``dwf_%s``.
+    Defaults to ``dwft_%s``.
```
jsocol, line 12: This is actually a different cookie than `dwft_`, this is the non-"testing" version.

jsocol:
#184 got the second typo, and the first isn't one. Thanks!

Issue #144, opened 2015-01-10
Move is_active logic out of __init__.py


jsocol:
Move most of the logic in `*_is_active` to `*.is_active` instance methods. It makes more sense for the model to contain this logic, since the logic codepends on the structure of the model and the state of the instance.

It also helps clean up `__init__.py`. The existing `*_is_active` functions can be thin wrappers that use the caching accessors (see #137).

PR #185, opened 2015-11-24
+ moved and create a user centric flag test method

slorg1:
This pull request is to allow "workers" and other non request-centric portion of the application to use waffle.

I basically just moved the code out of the single method and had the original method call the new.

### DIFF
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -56,39 +56,10 @@ def flag_is_active(request, flag_name):
         if tc in request.COOKIES:
             return request.COOKIES[tc] == 'True'
 
-    user = request.user
-
-    if flag.authenticated and user.is_authenticated():
-        return True
-
-    if flag.staff and user.is_staff:
-        return True
-
-    if flag.superusers and user.is_superuser:
-        return True
-
     if flag.languages:
         languages = flag.languages.split(',')
         if (hasattr(request, 'LANGUAGE_CODE') and
                 request.LANGUAGE_CODE in languages):
-            return True
-
-    flag_users = cache.get(keyfmt(get_setting('FLAG_USERS_CACHE_KEY'),
-                                              flag.name))
-    if flag_users is None:
-        flag_users = flag.users.all()
-        cache_flag(instance=flag)
-    if user in flag_users:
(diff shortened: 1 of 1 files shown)


jsocol:
I see what you're going for here, but I can't merge this right now. It conflicts with a bunch of things on the [current roadmap](https://github.com/jsocol/django-waffle/milestones)—and it won't be mergeable after those changes happen.

I filed #186 to track making this change in the future. By making an equivalent change after #144, we can avoid the duplication of the lookups (it stinks to have `flag_is_active()` look up the object then have `flag_is_active_for_user()` look it up again, and while that's avoidable, it introduces more complications to unwind later).

jsocol:
Thanks for the suggestion, though! It's an easy win and will be easier to incorporate further down the line.

slorg1:
Hi @jsocol,

Thank you for the feedback. For now I will use my fork (as I need the change). Let me know if I can help bringing it back later: may I introduce it in your latest branch?

I did think about the function call as well and my thoughts went as follows:
1. from the 1st call the cache is primed.
2. the second call from the 1st method allows centralized logic
3. the "extra" fetch actually hits the cache, so we are looking at: cost of cache hit and parse.

It seemed low enough and worth the extra cost.

jsocol:
It _probably_ hits the cache, but especially on cloud infra like elasticache or google's memcache, "write then immediately read" misses a lot more often than you'd hope. 

And even so, cache access is network access, so a few orders of magnitude slower than "self.active_for_user" or passing the Flag object in as an optional kwarg. 

##

slorg1:
True about the cache. There is always a risk of a miss with caching, though it is much less risky than using an instance variable with the risk of looking at the wrong user and jeopardizing multi threading. Functional programming works better in this context, I would suggest holding no states.
About the argument (kwarg): I thought of that too and was concerned about entering the risky road of enforcing mutual exclusion with the flag name. An other way was making 1 more method to allow a method taking the flag name and setup the one with the Flag as a delegate. I tried to keep my change small to keep your code clean.
I am curious to see the long term fix. Let me know if I can help. Thank you for the insights.

Issue #37, opened 2012-06-14
Add IP/subnet to flags


jsocol:
It would be valuable to turn on flags by internal IP or subnet. A comma-separated list of IP addresses with wild cards would be great.

adamhaney:
I currently have a need to be able to enable/disable flags by IP/subnet as well as the logic NOT of an IP/subnet and would be interested in contributing this. Does anyone have any strong opinions on API/design and/or would project maintainers be open to merging such functionality if I were to write it? This is probably a question best answered by @clintonb.

clintonb:
@adamhaney sorry for the delay. I have no strong opinions. I'll provide feedback on the PR.

PR #187, opened 2015-12-04
Add initial attempt at env-var based flag specification (and tests).

ntoll:
This PR introduces a new capability. **It is unfinished** (only flagging is supported) and this PR is where discussion of ongoing development can be had with upstream project owner. This work is being done with @rossbruniges.

**What?**

I'm working for a client whose Django website is "stateless" (i.e. does not have a database backend). This branch makes it easy to use three arbitrary "buckets" for specifying flags defined in environment variables: ALPHA, BETA and ALL, thus making it possible to use Waffle with websites that are not backed by a database.

**Why?**

My client wants to use feature flags and Waffle is awesome. Furthermore, not all websites are "typical" in that they use a database for storing state.

**How?**

There are three arbitrary buckets: ALPHA, BETA and ALL. ALPHA and BETA have named users associated with them and flags set for each bucket will only be active for those users. ALL related flags apply to all users. The choice of arbitrary bucket names was made to reflect the common pattern of rolling out a feature first in "alpha", then "beta" and then to "all" users. It is assumed that once a feature has been around long enough and there's no requirement for it to be on/off at short notice, then the flag related code should be removed from this feature.

We use the following environment variables for specifying flag/user state:
- `WAFFLE_USE_ENV_VARS` - a simple truthy flag to indicate that waffle should use the env vars instead of the database when checking if a `flag_is_active`.
- `WAFFLE_ALPHA_USERS` - a comma separated list of usernames for those in the ALPHA bucket (the stateless Django app will need to implement a faux-`request.user.username` to identify the individual users).
- `WAFFLE_BETA_USERS` - as with WAFFLE_ALPHA_USERS but for BETA users.
- `WAFFLE_ALPHA_FLAGS` - a comma separated list of flags in the ALPHA bucket.
- `WAFFLE_BETA_FLAGS` - a comma separated list of flags in the BETA bucket.
- `WAFFLE_ALL_FLAGS` - a comma separated list of flags in the ALL bucket.
- `WAFFLE_SWITCHES` - TBD.
- `WAFFLE_SAMPLES` - TBD.

This is most definitely a first draft of this functionality. I'm not precious about my code: comments, constructive critique and ideas are most welcome. "Not interested" is also a valid response! :-)

In any case, my client needs this feature so I'll be working on it over the coming weeks and, if useful, would like to make the work available upstream.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 # These are required to run the tests.
 Django
+django_jinja
 flake8
 
 # Everything else is in a Django-version-free version
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -2,12 +2,44 @@ from __future__ import unicode_literals
 
 from decimal import Decimal
 import random
+import os
 
 from waffle.utils import get_setting, keyfmt
 
 
 VERSION = (0, 11)
 __version__ = '.'.join(map(str, VERSION))
+
+
+def parse_env_vars(var_name):
+    """
+    Takes the name of an environment variable that should be expressed as a
+    comma separated list of items. Returns the equivalent Python data
+    structure.
+    """
+    value = os.getenv(var_name, '')
+    if value:
+        return [x for x in value.split(',') if x]
+    return []
+
+"""
+We default the settings derived from environment vars so the app can't break
+when it's started because we've forgotton to configure something. Rather, it'll
+just assume an 'all flags off' policy.
+
+There are three arbitrary buckets: ALPHA, BETA and ALL. ALPHA and BETA have
+named users associated with them and flags set for each bucket will only be
+active for those users. ALL related flags apply to all users.
+"""
+USE_ENV_VARS = os.getenv('WAFFLE_USE_ENV_VARS', False)
+if USE_ENV_VARS:
+    ALPHA_USERS = parse_env_vars('WAFFLE_ALPHA_USERS')
+    BETA_USERS = parse_env_vars('WAFFLE_BETA_USERS')
+    ALPHA_FLAGS = parse_env_vars('WAFFLE_ALPHA_FLAGS')
+    BETA_FLAGS = parse_env_vars('WAFFLE_BETA_FLAGS')
+    ALL_FLAGS = parse_env_vars('WAFFLE_ALL_FLAGS')
+    SWITCHES = parse_env_vars('WAFFLE_SWITCHES')
+    SAMPLES = parse_env_vars('WAFFLE_SAMPLES')
 
 
 class DoesNotExist(object):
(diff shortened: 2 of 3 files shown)


peterbe:
Why not leave the env var parsing to decouple and then use the django settings framework instead?

(hi! by the way @ntoll We meet again in django-land!)

jsocol:
@rossbruniges mentioned that y'all might be doing something like this, and I warned that I wasn't likely to merge it. With this in hand, I see a couple of significant issues that make it even less likely.
- Introducing "buckets" like this is a huge organizational shift. Waffle-with-env-vars would be a dramatically different project than Waffle-with-a-db.
- So is introducing a requirement to restart running processes to change flag state.
- So is identifying users by name instead of ID.
- There is no support for percentage rollouts, groups, or other user states, or a route toward supporting #3 or #37. Adding the parts that make sense would require a big investment in parallel code. which is a lot of maintenance to take on, and there would never be parity.

I think the best option is probably to start a new project or to fork Waffle. Waffle aims to support 90%/mainstream usecases, and "django without a database" doesn't fall into that. Plus, if someone did have a database but wanted to use env vars to configure feature flags, they'd still be able to via a new project.

A new project would be able to delete hundreds of lines of code here and expose the same sort of API in templates. The body of `switch_is_active` could be replaced with `return bool(int(os.getenv(name, 0)))`. `sample_is_active` would get similarly shorter.

Or, like @peterbe said, rely more heavily on the settings framework.

ntoll:
@jsocol quite understand - as you probably realise, much of the way this works depends on my client's "context". I think the best way forward is to fork into my client's organisation and they can run with it. Thanks for the suggestions too.

@peterbe hey hey matey. Yes, settings framework could work. I'll look into that. ;-)

jsocol:
> much of the way this works depends on my client's "context".

I'm also not keen to take on the maintenance cost of all this new code for a single client that—I'm guessing—isn't going to contribute resources to help keep Waffle running after this project. I'm going to close this out.

If you do fork it to a public place, all I ask is that it not be called "waffle". (I have no mechanism to enforce that, just saying "please.")

ntoll:
@jsocol ok I've forked it and renamed it "Waff" (as in Waffle, but without a back-end). For reference, it's here: https://github.com/ntoll/waff but will probably become re-owned by DigitalInnovation (part of Marks and Spencer - my client) when I've finished the work.

jsocol:
Cool. I'd really encourage you to drop all the database stuff and strip out those parts of the code, it'll be a lot easier to maintain.

ntoll:
Totally agree. That'll take some time, but the current "cut" of the code is usable.

PR #191, opened 2016-01-29
Easier Django upgrades by caching a plain list instead of `ValuesListQuerySet`

zacharyvoase:
`waffle.views._generate_waffle_js` was storing the result of `values_list` in a cache key. Although it looks like a list, it's actually an instance of `ValuesListQuerySet`. I encountered a bug in an application that was using a cache backend that pickled complex Python objects: upgrading major Django versions meant the values currently in the cache wouldn't deserialize correctly, throwing an error in the wafflejs view. Using vanilla lists of strings (which is really the intended effect here) makes it slightly more resistant to these upgrades.

I also fixed a brittle test case that broke on Django 1.9 (the `@never_cache` decorator now returns a few more directives).


jsocol:
Thanks!

PR #195, opened 2016-03-23
Add feature to activate flag by a cookie

alexcrawley:
This is motivated by wanting to trigger server side flag through conditions set though a 3rd party a/b testing service.


jsocol:
The right way to do this is to use [`waffle.set_flag`](https://github.com/jsocol/django-waffle/blob/master/waffle/__init__.py#L20) and set the flag based on the other cookie.

PR #196, opened 2016-04-07
Add python 3.5

nikolas:


jsocol:
I do need to add python3.5, but a few other things need to change here, too, especially dropping old versions. Thanks for getting it started!

PR #205, opened 2016-05-18
Update GA snippet

browniebroke:
In user testing the snippet is the legacy google analytics, updating to analytics.js syntax. Also add a link to Google analytics doc.


jsocol:
Thanks!

Issue #168, opened 2015-06-05
PicklingError on flag.users


ryanisnan:
We've been using Waffle for some time without issue, but recently upgraded to django 1.8 and waffle 0.10.1, and have started seeing a flood of these errors:

PicklingError: Can't pickle <class 'waffle.models.Flag_users'>: attribute lookup waffle.models.Flag_users failed

Kobold:
+1, I'm seeing this as well. Restarting the server seems to fix it temporarily. Here's a stack trace:

```
Traceback (most recent call last):
  File "/.../local/lib/python2.7/site-packages/django/core/handlers/base.py", line 108, in get_response
    response = middleware_method(request)
  File "./findoffice/middleware.py", line 9, in process_request
    if waffle.flag_is_active(request, 'multiple_markets'):
  File "/.../local/lib/python2.7/site-packages/waffle/__init__.py", line 41, in flag_is_active
    cache_flag(instance=flag)
  File "/.../local/lib/python2.7/site-packages/waffle/models.py", line 119, in cache_flag
    cache.add(keyfmt(settings.FLAG_USERS_CACHE_KEY, f.name), f.users.all())
  File "/.../local/lib/python2.7/site-packages/django/core/cache/backends/locmem.py", line 38, in add
    pickled = pickle.dumps(value, pickle.HIGHEST_PROTOCOL)
PicklingError: Can't pickle <class 'waffle.models.Flag_users'>: attribute lookup waffle.models.Flag_users failed
```

Django==1.8.2
django-waffle==0.10.1

Kobold:
It also seems that this is occurring only on flags that have users set on them.

TechNickAI:
We're experiencing the same issue

Django==1.8.1
django-waffle==0.10.1

TechNickAI:
As a workaround, I disabled caching for django waffle by setting `WAFFLE_CACHE_NAME` to a dummy cache backend.

Note: this will cause a lot of extra sql statements to your db. We use django-cacheops to cache at the sql level, so this layer of caching was redundant for us.

ryanisnan:
We're doing the same workaround @gorillamania. Anyone had further experience with this?

hakjoon:
I believe this was related to this bug with SimpleLazyObject which was fixed in 1.8.5 https://code.djangoproject.com/ticket/25389

I have not had this problem crop back up since upgrading to Django>=1.8.5

jsocol:
[...]

PR #211, opened 2016-05-28
Overhaul caching.

### LINKED ISSUE #77: wafflejs is (very) query intensive

### LINKED ISSUE #137: Stop using signals to cache objects

### LINKED ISSUE #158: uncache code is using the incorrect call
The uncache code is doing cache.set('name', None). This is specifically recommended against in the django docs. 

> We advise against storing the literal value None in the cache, because you won’t be able to distinguish between your stored None value and a cache miss signified by a return value of None.

It is causing strange behaviour in our system where the expected flag or switch value is not being returned under certain circumstances (e.g. in the unit tests where requests and updates happen very quickly)

The main unit tests are all passing but the override ones are not. I don't have sufficient understanding of what this is intended to do to fix them.  Shouldn't the update called by the override fire the uncache signal? This does not seem to be happening.

### LINKED ISSUE #163: If caching is working, test utils override_ methods break
**Currently caching does not work properly because nothing can ever be added to the cache, ([see the other issue I have made](https://github.com/jsocol/django-waffle/issues/162)).** If caching is fixed, the following issue arrises:

The update methods in the override_\* classes in testutils.py do not work when the model being overridden has been cached before the context manager that does the overriding is executed.

The updates() methods on the override_\* classes need to call save() on the model to clear the cache in order to guarantee that code inside the context manager that the override_\* method operates on does not use a cached version of the model. The current implementations do not do this this. eg:
`def update(self, active):
        self.cls.objects.filter(pk=self.obj.pk).update(everyone=active)`

According to the Django docs:
`Finally, realize that update() does an update at the SQL level and, thus, does not call any save() methods on your models, nor does it emit the pre_save or post_save signals (which are a consequence of calling Model.save()).`

The easiest way to fix this is to have the update() methods explicitly set the value and call save() on the model instance which will trigger the uncache method that has been connected in models.py. eg:
`def update(self, active):
        self.obj.everyone = active
        self.obj.save()`

This clears the cache and ensures that the first time the model being overridden is used inside the context manager, it will be retrieved from the database the first time it is accessed and contain the correct 'active value'.

I have made all these changes in the [master branch of our fork](https://github.com/festicket/django-waffle). Let me know if you would like me to make Pull Request for them.

### LINKED ISSUE #170: Cache flag_values and sample_values for _generate_waffle_js
May need to do https://github.com/jsocol/django-waffle/issues/162 first?

Calling `flag_is_active` and `sample_is_active` inside the list comprehension for every flag and sample starts to add up when there are a number of flags and samples to process. (73% of time spent in `_generate_waffle_js`)

Caching the `flag_values` and `sample_values` should help short-circuit that logic and make it perform better.

### LINKED ISSUE #186: "Flag.is_active_for_user" method
If we split the `is_active` method into its top and bottom halves—things that are focused on the `request` object and then things that are focused on the `request.user` object—as in #185, we can expose the bottom half as `is_active_for_user`. This needs to come after #144, though.

jsocol:
- Removes all signal handlers, fixes #137.
- Adds Cls.get_all() methods, fixes #142.
- Moves caching globally into Cls.get() and Cls.get_all(), fixes #170,
  fixes #160, fixes #77.
- Adds managers to invalidate ALL_\* cache keys on create.
- Adds a new delete admin action to invalidate cache.
- Adds Flag.is_active_for_user, fixes #186.
- Uses delete_many to flush cache, fixes #158.
- Fixes cache invalidation in override_*, fixes #163.

### DIFF
--- a/waffle/admin.py
+++ b/waffle/admin.py
@@ -21,8 +21,15 @@ def disable_for_all(ma, request, qs):
 disable_for_all.short_description = 'Disable selected flags for everyone.'
 
 
+def delete_flags(ma, request, qs):
+    # Iterate over all objects to cause cache invalidation.
+    for f in qs.all():
+        f.delete()
+delete_flags.short_description = 'Delete the selected flags.'
+
+
 class FlagAdmin(admin.ModelAdmin):
-    actions = [enable_for_all, disable_for_all]
+    actions = [enable_for_all, disable_for_all, delete_flags]
     date_hierarchy = 'created'
     list_display = ('name', 'note', 'everyone', 'percent', 'superusers',
                     'staff', 'authenticated', 'languages')
--- a/waffle/managers.py
+++ b/waffle/managers.py
@@ -1,0 +1,30 @@
+from __future__ import absolute_import, unicode_literals
+
+from django.db import models
+
+from waffle.utils import get_setting, get_cache
+
+
+cache = get_cache()
+
+
+class BaseManager(models.Manager):
+    KEY_SETTING = ''
+
+    def create(self, *args, **kwargs):
+        ret = super(BaseManager, self).create(*args, **kwargs)
+        cache_key = get_setting(self.KEY_SETTING)
+        cache.delete(cache_key)
+        return ret
+
+
+class FlagManager(BaseManager):
+    KEY_SETTING = 'ALL_FLAGS_CACHE_KEY'
+
+
+class SwitchManager(BaseManager):
+    KEY_SETTING = 'ALL_SWITCHES_CACHE_KEY'
+
+
+class SampleManager(BaseManager):
+    KEY_SETTING = 'ALL_SAMPLES_CACHE_KEY'
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -14,11 +14,81 @@ from django.db import models
 from django.db.models.signals import post_save, post_delete, m2m_changed
 from django.utils.encoding import python_2_unicode_compatible
 
+from waffle import managers
 from waffle.utils import get_setting, keyfmt, get_cache
 
 
 cache = get_cache()
 CACHE_EMPTY = '-'
+
+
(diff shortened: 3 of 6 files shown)


groovecoder:
🎉

joeyespo:
Just ran into #168. Any chance this will be released soon?

PR #218, opened 2016-06-04
Flag as an abstract base class

cleder:


FlogFr:
Excellent work/contribution!
I tried to think twice and do a PR on possible default `Flag` model inside `django-waffle`, and I think you're right on the explicit over implicit. It would be weird to have a default `Flag` model which would be not represented in the migrations and then the migrations would need to be generated in the `django-waffle` applications...

Thanks again :)

cleder:
@jsocol any thoughts about this PR?

jsocol:
I haven't had a lot of time to dig into this one yet, but I have a hard time seeing myself merging it as-is. I think there are some good ideas here, but they're wound up with a bunch of other changes.

I've been underwater with work this month, but I'm going to try to find some time to keep moving on 0.12 over the holiday weekend. I don't expect abstract base models to be part of that release.

koliber:
I really like the way Django implements custom user models. I recently had an opportunity to ask Russell Keith-Magee, the author of Django's custom user models, on what he would recommend. He explained how that was done and pointed out some pitfalls. 

I took his advice and build custom flag models emulating the way Django did custom user models: https://github.com/jsocol/django-waffle/pull/243 .

PR #225, opened 2016-10-06
Fix tox config

adamchainz:
Broken by excessive deletions in 1a24a2ab9ab6311e0acb64707695feaab18c82c3 :D


jsocol:
I think this is subsumed by #227 / #239, so going to close in favor of those

PR #226, opened 2016-10-06
Use strings for cache keys

adamchainz:
As discovered in YPlan/django-perf-rec#21, django-waffle is using `bytes` objects for cache keys. This seems to work with most of the caching infrastructure, but it's not documented as supported in the [Django cache docs](https://docs.djangoproject.com/en/1.10/topics/cache/) nor does it seem to be tested for in its test suite. Thus some cache backends may not support `bytes`, and `django-perf-rec` currently doesn't support them either.

### DIFF
--- a/waffle/utils.py
+++ b/waffle/utils.py
@@ -22,7 +22,7 @@ def keyfmt(k, v=None):
         key = prefix + k
     else:
         key = prefix + hashlib.md5((k % v).encode('utf-8')).hexdigest()
-    return key.encode('utf-8')
+    return key
 
 
 def get_cache():
--- a/docs/starting/configuring.rst
+++ b/docs/starting/configuring.rst
@@ -41,7 +41,8 @@ behavior.
     Waffle tries to store objects in cache pretty aggressively. If you
     ever upgrade and change the shape of the objects (for example
     upgrading from <0.7.5 to >0.7.5) you'll want to set this to
-    something other than ``'waffle:'``.
+    something other than ``'waffle:'``. If you're using memcached this should
+    be ASCII only, as that's all it supports.
 
 ``WAFFLE_CACHE_NAME``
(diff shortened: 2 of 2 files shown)


waffle/utils.py
```diff
@@ -22,7 +22,7 @@ def keyfmt(k, v=None):
         key = prefix + k
     else:
         key = prefix + hashlib.md5((k % v).encode('utf-8')).hexdigest()
-    return key.encode('utf-8')
```
jsocol, line 25: Hmm, this was added specifically because unencoded utf8 characters break multiple cache backends (in particular memcache). The Python code is OK with it, but memcached itself was not. If the `CACHE_PREFIX` setting is set to an unsafe string, this will break. `k` should only come from within the library, so that should be OK, and `v` gets encoded and digested, so it should be safe.

At the very least, this would need docs added to the settings reference that made it clear `CACHE_PREFIX` needs to be plain ascii.

jsocol (changes requested):

waffle/utils.py
adamchainz: The plain ASCII restrict exists on memcache but afaik all other Django core cache backends support unicode, or can support it in the right conditions, e.g. correct database setup for dbcache. I can document it, but Django *does* already emit warnings when forming cache keys that won't work on memcached: https://github.com/django/django/blob/master/django/core/cache/backends/base.py#L226

adamchainz:
Added documentation

jeffbr13:
Hi there! Just wanted to note that I'm having the same issue using `django-perf-rec` as well! Since there are new maintainers (@rlucioni and @ahsan-ul-haq) perhaps we could rescue this from stale issues, since it's a simple (although potentially breaking) fix?

clintonb (approved):

PR #227, opened 2016-10-06
Travis - just run tox

adamchainz:
Just run `tox` when on travis, meaning only one grid of Python versions versus Django versions needs be maintained, plus `tox` won't break whilst the build is 'passing'.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,9 @@
 language: python
 sudo: false
-python:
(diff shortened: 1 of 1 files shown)


jsocol:
Fixed by #239

jsocol:
Thanks @adamchainz! This patch ended up in the other PR and was very helpful

Issue #182, opened 2015-10-21
A single settings dict


jsocol:
The trend in django core has been to move from lots of `NAMESPACE_*` settings to `NAMESPACE = {'FOO': 1}`. I think in the interest of reducing settings namespace bloat, and simplifying some code in a few places (well, replacing `getattr()` with `.get()`, mostly), this is a good trend and we should move the `WAFFLE_*` settings into a `WAFFLE` dict.

willkg:
I'm +1 on this.

jsocol:
Since it's another breaking change, I'm putting it in 0.12 for now.

PR #230, opened 2016-10-25
A single settings dict #182

cleder:
The trend in django core has been to move from lots of `NAMESPACE_*` settings to 
`NAMESPACE = {'FOO': 1}`. 
In the interest of reducing settings namespace bloat move the `WAFFLE_*` settings into a `WAFFLE` dict.


jsocol:
This is definitely something I want to do (see #182) but this PR covers a lot of different functionality, so I'm not going to merge it as-is. If you want to pull out the single settings dict as a first step, I'd be happy to look at it individually.

PR #236, opened 2016-11-11
Fixing waffle commands to run with django 1.10.3

jdavidagudelo:


jsocol:
Would these still run with 1.8 and 1.9? I realize there's no actual test coverage for the management commands. It'd be great to add something, even if it was just a smoke "yes you can execute these" kind of test.

chris-griffin:
Just giving this PR a bump. [`add_arguments()` was added in django 1.8](https://docs.djangoproject.com/en/1.10/releases/1.8/#management-commands-that-only-accept-positional-arguments) so it will work in 1.8 and 1.9. Could we do a release after this is merged to get django-waffle working with django 1.10 and 1.11?

waffle/management/commands/waffle_flag.py
```diff
@@ -1,65 +1,64 @@
...
-            default=False,
-            help='Turn on rollout mode.'),
-        make_option('--create',
-            action='store_true',
-            dest='create',
-            default=False,
-            help='If the flag doesn\'t exist, create it.'),
-    )
+    def add_arguments(self, parser):
+        percent_help = '{0} {1}'.format('Roll out the flag for a certain percentage of users. ',
```
clintonb, line 10: You don't need `str.format()` here. Just use a string constant.

waffle/management/commands/waffle_flag.py
```diff
@@ -94,4 +93,4 @@ def handle(self, flag_name=None, *args, **options):
                 print('Setting %s: %s' % (option, options[option]))
                 setattr(flag, option, options[option])
 
-        flag.save()
+        flag.save()
```
clintonb, line 96: Add a newline.

clintonb (approved):
Nit: Squash the commits, unless there is a good reason for this to be spread across multiple commits.

clintonb:
@jsocol @jdavidagudelo can we get this merged soon? We at edX need to get to 1.11.x since it is the current LTS.

phillbaker:
Looks like this was fixed by 177847a and can be closed?

PR #239, opened 2017-01-16
Add Django 1.10.x test environment and fix broken tests

### LINKED ISSUE #23: WAFFLE_SWITCH_DEFAULT and WAFFLE_FLAG_DEFAULT have no effect
```
WAFFLE_FLAG_DEFAULT = True
```

...in settings does not make this return True as expected:

```
waffle.flag_is_active('non-existant')
```

The same applies to template helpers and switches.

### LINKED ISSUE #224: Test against Django 1.10

djm:
Hi James,

Just a small one here to get the project's test suite up to current speed.

Tox now runs the test suite successfully under the latest Django 1.10.x and Python 3.6 environments.

As part of making the tests run under the new environement, the tests were updated to account for deprecated behaviour in Django. Only thing of note really is that `render_to_response` was removed in favour of `render` as `render_to_response` is ['likely pending deprecation'](https://docs.djangoproject.com/en/1.10/topics/http/shortcuts/#render-to-response) as of Django 1.10.

Thanks for all your work on this, we use it extensively at Yunojuno. If the 1.10 stuff could get a release on PyPI that'd be super; if there's anything blocking that just let us know.

Fixes #224.

cheers,
Darian

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ dist
 .project
 .pydevproject
 .settings
+.tox/
 docs/_*
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,7 @@
 language: python
 sudo: false
-python:
- - "2.7"
- - "3.3"
- - "3.4"
- - "3.5"
- - "pypy"
-env:
- - DJANGO_VERSION="1.8"
- - DJANGO_VERSION="1.9"
+python: "3.6"
 install:
- - pip install -q "Django>=${DJANGO_VERSION},<${DJANGO_VERSION}.99" -r travis.txt
-script: ./run.sh test
-matrix:
-  exclude:
-    - python: "3.3"
-      env: DJANGO_VERSION="1.9"
+ - pip install tox
+script:
+ - tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,16 @@
 language: python
+
 sudo: false
-python: "3.6"
+
+python:
+  - "2.7"
+  - "3.3"
+  - "3.4"
+  - "3.5"
+  - "3.6"
+
 install:
- - pip install tox
+ - pip install tox tox-travis
+
 script:
  - tox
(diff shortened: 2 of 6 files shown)


jsocol:
Hey @djm, thanks! Waffle doesn't use tox on travis, though, it uses the `.travis.yml` file to configure the jobs, so this didn't actually run the tests against 1.10. Everything ran OK against 1.8 and 1.9, though. Would you mind adding 1.10 to `.travis.yml`, too?

jsocol:
Alternatively, if you wanted to incorporate @adamchainz's patch from #227, that would also work

djm:
Ah! My bad assumption.

I'll do that ASAP, check it works - then squash.

Thanks @jsocol.

jsocol:
Looks like we need to exclude py3.3 with >=1.9?

One downside of tox is no parallel tests / individual combination reporting from the job, ah well.

jsocol:
Hmm, I guess this is still open for Py3.5 https://github.com/travis-ci/travis-ci/issues/4794

adamchainz:
> One downside of tox is no parallel tests / individual combination reporting from the job, ah well.

Tests being parallel isn't always faster, for many repos I've moved to having the grid in only tox it has ended up nearly the same time or even faster overall. You can do the coverage combining somehow in the tox file.

jsocol:
> Tests being parallel isn't always faster

Faster isn't the only thing (actually I think this run was faster than most) but when you use the Travis matrix you get the slightly cleaner report of whether each combination passed or failed, e.g. [the last master build](https://travis-ci.org/jsocol/django-waffle/builds/192440021). But like I said, "ah well," not worth maintaining two grids.

djm:
Ok, so this works, and keeps the builds separate but not repeated. [tox-travis](https://github.com/ryanhiebert/tox-travis) seems pretty handy.

If you're happy with it @jsocol, I can squash. If not, let me know! Thanks again.

jsocol:
This lgtm, thanks @djm!

djm:
Awesome, thanks @jsocol :)

jsocol:
"Squash and merge" is a pretty nice feature. Thanks again @djm!

PR #242, opened 2017-03-26
Get rid of reverse import DeprecationWarning 

coagulant:
Test against Django 1.11rc1
Forward-compatible django.urls import
Remove StrictVersion to allow support 1.11rc1, not-compatible with distutils


coagulant:
Fixed in #247 (partially)

PR #247, opened 2017-06-27
Added support for Django 1.11 and Python 3.6

clintonb:


clintonb:
@rlucioni please review

schenedx (approved):
👍 Do you think the two commits are useful being separate?

rlucioni (approved):

clintonb:
Leaving as two separate commits, since they are two independent changes.

PR #251, opened 2017-08-22
check request before using request. 

### LINKED ISSUE #250: error while rendering page without request context
while I use a flag with testing mode. 

and if I render a page without request context.

the error happen.

```
    def is_active(self, request):
        if not self.pk:
            return get_setting('FLAG_DEFAULT')

        if get_setting('OVERRIDE'):
            if self.name in request.GET:
                return request.GET[self.name] == '1'

        if self.everyone:
            return True
        elif self.everyone is False:
            return False

        if self.testing:  # Testing mode is on.
            tc = get_setting('TEST_COOKIE') % self.name
>           if tc in request.GET:
E           AttributeError: 'NoneType' object has no attribute 'GET'

```

timtan:
to fix #250 

please let me know if there are any information needed.

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -221,7 +221,7 @@ class Flag(BaseModel):
             return get_setting('FLAG_DEFAULT')
 
         if get_setting('OVERRIDE'):
-            if self.name in request.GET:
+            if request and self.name in request.GET:
                 return request.GET[self.name] == '1'
 
         if self.everyone:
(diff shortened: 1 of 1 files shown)


jsocol:
Thanks for the patch! But please see my comment https://github.com/jsocol/django-waffle/issues/250#issuecomment-324006426 on the issue. Flags should require requests.

PR #252, opened 2017-08-26
Add groups to waffle_flag management command

jwarren116:
This solves issue https://github.com/jsocol/django-waffle/issues/222.

This updates the `waffle_flag` management command to take multiple `--groups` or `-g` arguments for attaching Django permission groups to a Waffle flag. By default it will clear existing groups attached to the flag. However, this also adds an `--append` flag, which will alternatively append groups to the Waffle flag.

Additionally I updated a few flake8 errors, and added a test for the command.

### DIFF
--- a/waffle/management/commands/waffle_flag.py
+++ b/waffle/management/commands/waffle_flag.py
@@ -1,5 +1,6 @@
 from __future__ import print_function
 
+from django.contrib.auth.models import Group
 from django.core.management.base import BaseCommand, CommandError
 
 from waffle.models import Flag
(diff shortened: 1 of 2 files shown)


waffle/management/commands/waffle_flag.py
```diff
@@ -29,7 +30,8 @@ def add_arguments(self, parser):
             action='store',
             type=int,
             dest='percent',
-            help='Roll out the flag for a certain percentage of users. Takes a number between 0.0 and 100.0'),
+            help='Roll out the flag for a certain percentage of users. Takes '
```
jsocol, line 33: Thanks for fixing all these up!

jsocol (approved):

waffle/management/commands/waffle_flag.py
jwarren116: For sure! I noticed there are other Flake8 errors throughout the code. I'll open a PR to fix these in the next week or so.

PR #254, opened 2017-08-28
Decorating entire views and redirecting with args and kwargs

Belco90:
When using `waffle_flag` and `waffle_switch` to decorate entire views is not possible to redirect properly if views expect args or kwargs: the decorators are receiving the propers args and kwargs, but they are not passed to `get_response_to_redirect` in order to build the reverse url properly, so instead of get a 302 response, you get a 404 response.

I have modified proper code in `waffle/decorators.py` to pass args and kwargs from each decorator to `get_response_to_redirect` and then I have used them in `reverse` method, including corresponding tests.

I found this problem when creating a new page to edit some models but maintaining old page version (I need the PK instance in both view kwargs).


jsocol (approved):
Sorry I didn't take a look at this yet, but it looks good! Thanks!

Belco90:
Nice! Thanks.

PR #255, opened 2017-08-29
Fix Flake8 errors

jwarren116:
This fixes many of the previous flake8 errors that were present. A few `# noqa` lines were added where they make sense. I did not remove any of the unused imports, as I wasn't entirely sure if they were there for a good reason I'm not aware of. That being said, I'm happy to remove them if they aren't there for a good reason (or add a `# noqa` to exclude them from flake8 if they are needed). The remaining errors are below.

```
waffle/__init__.py:3:1: F401 'decimal.Decimal' imported but unused
waffle/__init__.py:4:1: F401 'random' imported but unused
waffle/__init__.py:6:1: F401 'waffle.utils.keyfmt' imported but unused
waffle/__init__.py:6:1: F401 'waffle.utils.get_cache' imported but unused
waffle/__init__.py:6:1: F401 'waffle.utils.get_setting' imported but unused
waffle/models.py:14:1: F401 'django.db.models.signals.m2m_changed' imported but unused
waffle/models.py:14:1: F401 'django.db.models.signals.post_save' imported but unused
waffle/models.py:14:1: F401 'django.db.models.signals.post_delete' imported but unused
waffle/views.py:7:1: F401 'waffle.sample_is_active' imported but unused
waffle/views.py:7:1: F401 'waffle.flag_is_active' imported but unused
waffle/views.py:9:1: F401 'waffle.utils.keyfmt' imported but unused
```

### DIFF
--- a/waffle/admin.py
+++ b/waffle/admin.py
@@ -18,7 +18,6 @@ def enable_for_all(ma, request, qs):
     for f in qs.all():
         f.everyone = True
         f.save()
-enable_for_all.short_description = 'Enable selected flags for everyone.'
 
 
 def disable_for_all(ma, request, qs):
@@ -26,13 +25,16 @@ def disable_for_all(ma, request, qs):
     for f in qs.all():
         f.everyone = False
         f.save()
-disable_for_all.short_description = 'Disable selected flags for everyone.'
 
 
 def delete_individually(ma, request, qs):
     # Iterate over all objects to cause cache invalidation.
     for f in qs.all():
         f.delete()
+
+
+enable_for_all.short_description = 'Enable selected flags for everyone.'
+disable_for_all.short_description = 'Disable selected flags for everyone.'
 delete_individually.short_description = 'Delete selected.'
 
 
@@ -49,13 +51,15 @@ def enable_switches(ma, request, qs):
     for switch in qs:
         switch.active = True
         switch.save()
-enable_switches.short_description = 'Enable the selected switches.'
 
 
 def disable_switches(ma, request, qs):
     for switch in qs:
         switch.active = False
         switch.save()
+
+
+enable_switches.short_description = 'Enable the selected switches.'
 disable_switches.short_description = 'Disable the selected switches.'
 
 
--- a/waffle/tests/test_templates.py
+++ b/waffle/tests/test_templates.py
@@ -35,8 +35,15 @@ class WaffleTemplateTests(TestCase):
         self.assertContains(response, 'window.waffle =')
 
     def test_get_nodes_by_type(self):
-        """WaffleNode.get_nodes_by_type() should correctly find all child nodes"""
-        test_template = Template('{% load waffle_tags %}{% switch "x" %}{{ a }}{% else %}{{ b }}{% endswitch %}')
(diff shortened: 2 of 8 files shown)


waffle/admin.py
```diff
@@ -18,22 +18,22 @@ def enable_for_all(ma, request, qs):
     for f in qs.all():
         f.everyone = True
         f.save()
-enable_for_all.short_description = 'Enable selected flags for everyone.'
+enable_for_all.short_description = 'Enable selected flags for everyone.'  # noqa: E305,E501
```
jsocol: Is there a way to fix these without opting out of the linter? E305 is missing blank lines and E501 is line too long, right?

waffle/templatetags/waffle_tags.py
```diff
@@ -70,13 +70,13 @@ def flag(parser, token):
 
 @register.tag
 def switch(parser, token):
-    condition = lambda request, name: switch_is_active(name)
+    condition = lambda request, name: switch_is_active(name)  # noqa: E731
```
jsocol: is E731 no lambdas? we could swap these to be non-lambdas or allow lambdas in the setup.cfg config

waffle/tests/test_templates.py
```diff
@@ -35,8 +35,11 @@ def test_django_tags(self):
         self.assertContains(response, 'window.waffle =')
 
     def test_get_nodes_by_type(self):
-        """WaffleNode.get_nodes_by_type() should correctly find all child nodes"""
-        test_template = Template('{% load waffle_tags %}{% switch "x" %}{{ a }}{% else %}{{ b }}{% endswitch %}')
+        """WaffleNode.get_nodes_by_type() should find all child nodes."""
+        test_template = Template(
+            '{% load waffle_tags %}{% switch "x" %}{{ a }}'
```
jsocol: nit but can you break this into a few lines? i think it'll be a _lot_ more readable

```py
'{% load waffle_tags %}'
'{% switch "x" %}'
'... etc ...'
```

waffle/tests/test_waffle.py
```diff
@@ -298,7 +298,8 @@ def test_switch_active_from_cache(self):
         assert waffle.switch_is_active(switch.name)
         queries = len(connection.queries)
         assert waffle.switch_is_active(switch.name)
-        self.assertEqual(queries, len(connection.queries), 'We should only make one query.')
+        self.assertEqual(queries, len(connection.queries),
```
jsocol: tbh these messages probably aren't even that helpful, we could drop them

jsocol (commented):
thanks for this! a couple small comments or questions—but I'd really like to avoid `#noqa` as a solution where possible

jwarren116:
I've pushed updates based on the comments.

One note: flake8 wants two lines before/after functions or classes, which would add up to a lot of white space in the `admin.py`. So I grouped the descriptions above the Admin class for each model. I like the visual organization of the description immediately after the function, but that would require the `# noqa` or allowing it in the `setup.cfg` (which would allow it everywhere, which isn't my favorite idea). I think grouping them is the best alternative.

jsocol (approved):
Thank you! This looks excellent

PR #256, opened 2017-09-04
Cleanup and fix management commands.

vil-s:
- Ensure that new switch is not created unless the --create flag is passed.
- Add tests to ensure no regression for the --create fix on all commands.
- Add separate positional arguments for commands (instead of "positionals") to improve the --help documentation.

### DIFF
--- a/waffle/management/commands/waffle_flag.py
+++ b/waffle/management/commands/waffle_flag.py
@@ -8,74 +8,76 @@ from waffle.models import Flag
 
 class Command(BaseCommand):
     def add_arguments(self, parser):
-        parser.add_argument('name', nargs='?')
+        parser.add_argument(
+            'name',
+            nargs='?',
+            help='The name of the flag.')
         parser.add_argument(
             '-l', '--list',
             action='store_true',
             dest='list_flags',
             default=False,
-            help="List existing samples."),
+            help='List existing samples.')
         parser.add_argument(
             '--everyone',
             action='store_true',
             dest='everyone',
-            help="Activate flag for all users."),
+            help='Activate flag for all users.')
         parser.add_argument(
             '--deactivate',
             action='store_false',
             dest='everyone',
-            help="Deactivate flag for all users."),
+            help='Deactivate flag for all users.')
         parser.add_argument(
             '--percent', '-p',
             action='store',
             type=int,
             dest='percent',
             help='Roll out the flag for a certain percentage of users. Takes '
-                 'a number between 0.0 and 100.0'),
+                 'a number between 0.0 and 100.0')
         parser.add_argument(
             '--superusers',
             action='store_true',
             dest='superusers',
             default=False,
-            help='Turn on the flag for Django superusers.'),
+            help='Turn on the flag for Django superusers.')
         parser.add_argument(
             '--staff',
             action='store_true',
             dest='staff',
             default=False,
-            help='Turn on the flag for Django staff.'),
+            help='Turn on the flag for Django staff.')
         parser.add_argument(
             '--authenticated',
             action='store_true',
             dest='authenticated',
             default=False,
-            help='Turn on the flag for logged in users.'),
+            help='Turn on the flag for logged in users.')
         parser.add_argument(
             '--group', '-g',
             action='append',
             default=list(),
-            # dest='group_name',
             help='Turn on the flag for listed group names (use flag more '
                  'than once for multiple groups). WARNING: This will remove '
-                 'any currently associated groups unless --append is used!'),
+                 'any currently associated groups unless --append is used!')
         parser.add_argument(
             '--append',
             action='store_true',
             dest='append',
             default=False,
-            help='Append only mode when adding groups.'),
+            help='Append only mode when adding groups.')
         parser.add_argument(
             '--rollout', '-r',
             action='store_true',
             dest='rollout',
             default=False,
-            help='Turn on rollout mode.'),
+            help='Turn on rollout mode.')
         parser.add_argument(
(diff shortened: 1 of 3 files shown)


waffle/management/commands/waffle_flag.py
```diff
@@ -8,74 +8,76 @@
...
+            'name',
+            nargs='?',
+            help='The name of the flag.')
         parser.add_argument(
             '-l', '--list',
             action='store_true',
             dest='list_flags',
             default=False,
-            help="List existing samples."),
+            help="List existing samples.")
```
clintonb: Single quotes

waffle/management/commands/waffle_flag.py
```diff
@@ -8,74 +8,76 @@
...
         parser.add_argument(
             '--everyone',
             action='store_true',
             dest='everyone',
-            help="Activate flag for all users."),
+            help="Activate flag for all users.")
```
clintonb: Single quotes

waffle/management/commands/waffle_flag.py
```diff
@@ -8,74 +8,76 @@
...
         parser.add_argument(
             '--deactivate',
             action='store_false',
             dest='everyone',
-            help="Deactivate flag for all users."),
+            help="Deactivate flag for all users.")
```
clintonb: Single quotes

waffle/management/commands/waffle_sample.py
```diff
@@ -51,7 +60,7 @@ def handle(self, *args, **options):
             try:
                 sample = Sample.objects.get(name=sample_name)
             except Sample.DoesNotExist:
-                raise CommandError('This sample does not exist.')
+                raise CommandError("This sample doesn't exist.")
```
clintonb: Single quotes. I find "does not" to be more understandable (esp. for non-native English speakers) than "doesn't".

waffle/tests/test_management.py
```diff
@@ -28,6 +28,15 @@ def test_create(self):
         self.assertEqual(list(flag.groups.values_list('name', flat=True)),
                          ['waffle_group'])
 
+    def test_not_create(self):
+        """ The command shouldn't create a new flag if the create flag is
+        not set.
+        """
+        with self.assertRaisesRegexp(CommandError, "This flag doesn't exist."):
```
clintonb: Applicable to all of your new tests:

1. Use single quotes to maintain consistency.
2. Add an assertion that actually ensures the flag/switch/sample does not exist.

clintonb (changes requested):
Please squash your commits.

vil-s:
Review changes implemented

waffle/tests/test_management.py
```diff
@@ -28,6 +28,17 @@ def test_create(self):
         self.assertEqual(list(flag.groups.values_list('name', flat=True)),
                          ['waffle_group'])
 
+    def test_not_create(self):
+        """ The command shouldn't create a new flag if the create flag is
+        not set.
+        """
+        name = 'test'
+        self.assertFalse(Flag.objects.filter(name=name).exists())
```
clintonb: The assertions should come _after_ the calls to the management command. The purpose of the test is to ensure the command doesn't create a new instance.

clintonb (changes requested):
Correct the tests.

waffle/tests/test_management.py
vil-s: aha, I misunderstood your previous comment

waffle/tests/test_management.py
jsocol: Fwiw, I strongly prefer double quotes when the string has apostrophes in it, to avoid awkward escape sequences:

```py
'don\'t do this if it can be avoided'
"this's better"
'but this is best'
```

codeadict:
This will be a very nice to have.

vil-s:
@clintonb I fixed the tests and fixed the merge conflicts. It should be ready to go now.

clintonb (approved):

clintonb:
@unklphil please rebase on the master branch, rather than merge master into your own branch. Once that is done, I'll merge.

aehlke:
Just curious why rebasing/squashing in PRs is critical now that github allows squashing as an option in the merge button?

clintonb:
This project using the "Rebase and merge" option solely. That decision for this project wasn't mine, but I agree with it. Merging introduces merge bubbles which are mostly noise. Squashing assumes that every PR should be squashed into a single commit when. The decision to use a single commit is not always applicable. If I happen to fix two separate bugs, they should be two separate commits.

In this case, since "Rebase and merge" is the only option available to me, the branch needs to be properly rebased. GitHub will not allow the PR to be merged without a rebase.

screen shot 2017-10-24 at 11 09 58 pm

aehlke:
Thank you, I appreciate the detailed response - I've wondered about this, that makes sense now!

vil-s:
I used the wrong damned option for resolving management.commands.waffle_switch after squashing and rebasing (I was certain that "use mine" would use mine, but it didn't) and did `git push --force`, and now those changes are lost.

I really don't feel like fixing it, so I'm going to close the issue this PR because it's too much of a hassle for me.

jsocol:
Ugh, I'm sorry, that's the worst!

If you haven't deleted the local checkout, it may be possible to recover, if you want—or for a future situation—via the `git reflog`. It can fix things like accidental rebases, recover things that have been forced-pushed over, etc. There be dragons but it's a very powerful tool.

vil-s:
I'll check tomorrow if it's salvagable

PR #259, opened 2017-10-20
user.is_authenticated is not callable anymore in django>=1.10

Albyorix:
For reference, here is the change https://github.com/django/django/commit/c1aec0feda73ede09503192a66f973598aef901d#diff-0525989eb70bcc9e943110ba53a6dad6

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -190,7 +190,10 @@ class Flag(BaseModel):
         return group_ids
 
     def is_active_for_user(self, user):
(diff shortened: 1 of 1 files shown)


clintonb (commented):
This looks good, but needs a test case. Also, please squash to a single commit.

ojongerius:
@Albyorix do you have time implement @clintonbs suggestions? I need this too.

clintonb:
See #266

PR #266, opened 2017-12-14
Updated User.is_authenticated usage to be compatible with Django 1.10+

clintonb:
User.is_authenticated is now a property in Django 1.10+ rather than a method. This change updates the usage to account for this change, while retaining backward compatibility with older versions of Django.

### DIFF
--- a/waffle/tests/test_waffle.py
+++ b/waffle/tests/test_waffle.py
@@ -161,7 +161,6 @@ class WaffleTests(TestCase):
         assert 'dwf_myflag' not in response.cookies
 
         request.user = User(username='foo')
-        assert request.user.is_authenticated()
         response = process_request(request, views.flag_in_view)
         self.assertEqual(b'on', response.content)
         assert 'dwf_myflag' not in response.cookies
@@ -177,7 +176,6 @@ class WaffleTests(TestCase):
         assert 'dwf_myflag' not in response.cookies
 
         request.user = User(username='foo')
-        assert request.user.is_authenticated()
         response = process_request(request, views.flag_in_view)
         self.assertEqual(b'on', response.content)
         assert 'dwf_myflag' not in response.cookies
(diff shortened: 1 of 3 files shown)


clintonb:
Replaces #259

waffle/tests/test_waffle.py
```diff
@@ -161,7 +161,6 @@ def test_authenticated(self):
         assert 'dwf_myflag' not in response.cookies
 
         request.user = User(username='foo')
-        assert request.user.is_authenticated()
```
KostyaEsmukov, line 164: I would suggest to define a function at the top of the file (or maybe somewhere in utils), like [raven does](https://github.com/getsentry/raven-python/blob/4c59089f6d1fbe5df1c7b05a8ed5d284870b6c2e/raven/contrib/django/client.py#L41):

	if DJANGO_VERSION < (1, 10):
	    def is_authenticated(request_user):
	        return request_user.is_authenticated()
	else:
	    def is_authenticated(request_user):
	        return request_user.is_authenticated

and then use it like this:

    assert is_authenticated(request.user)

Also I guess that testing by callability of `is_authenticated` instead of the Django version still causes the warning with Django 1.10.

jsocol: I always prefer feature testing to version number testing, something like `if callable(request.user.is_authenticated)`

KostyaEsmukov: Yeah, me too, but the problem is that with Django 1.10-1.11 one would still see the warning when using the callable version of this property. Though I didn't check this guess.

jsocol: Deprecation warnings are usually only emitted once per process so I'm not too worried about that. But #267 may eliminate this issue anyway.

D3X:
@jsocol I think this can be closed now that #269 was merged.

jsocol:
Yep, thanks!

PR #267, opened 2017-12-14
Added support for Django 2.0

clintonb:


jsocol:
What would it take to support 1.8, 1.11, and 2.0? Those are the actively supported release lines right now and I've always committed to supporting ESRs. (On the other hand, 1.10 has reached EOL so I'd drop that.) Before dropping an active ESR I want to really understand how costly it would be to keep.

clintonb:
Supporting releases prior to Django 1.10 will require more finagling than it's worth given that (a) 1.8 support goes away after April 2018 and (b) users can still rely on older releases. This package seems pretty stable, so I don't foresee any issues with people using the older versions until they upgrade to Django 2.0.

The [Django roadmap](https://www.djangoproject.com/weblog/2015/jun/25/roadmap/) states that "apps may drop support for Django 1.8, 1.9, and 1.10," now that 2.0 is released.

D3X:
@jsocol @clintonb I gave it a try and it wasn't as hard as it seemed to support everything from 1.8 to 2.0
See: #269

D3X:
@jsocol I think this can be closed now that #269 was merged.

jsocol:
Agreed, thanks!

PR #268, opened 2017-12-15
Feature: more granular querystring flag overrides

slafs:
This enables the user to choose individual flags to be overridden in
the querystring. Currently we have a binary choice (all or none) via
`WAFFLE_OVERRIDE` setting.

Adds a new `NullBooleanField` field on the `Flag` model called
`allow_override`. The default value (`None`) means that the control
depends on the global `WAFFLE_OVERRIDE` value. When set to `False`
disables the override even if `WAFFLE_OVERRIDE` is set to `True`.
Finally, when the field is set to `True` we allow querystring
overrides for this flag (as described in documentation) even if
`WAFFLE_OVERRIDE` is set to `False`.

### DIFF
--- a/docs/starting/configuring.rst
+++ b/docs/starting/configuring.rst
@@ -40,8 +40,9 @@ behavior.
     Switches, and Samples from the DB configured for writes on cache misses.
 
 ``WAFFLE_OVERRIDE``
-    Allow *all* Flags to be controlled via the querystring (to allow
-    e.g. Selenium to control their behavior). Defaults to ``False``.
+    Allow Flags to be controlled via the querystring (to allow e.g. Selenium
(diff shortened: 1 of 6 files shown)


jsocol:
Is there a reason the [`testing` boolean](http://waffle.readthedocs.io/en/stable/testing/user.html#small-scale-tests) doesn't already cover this?

slafs:
Well... Looks like it does... I'm not sure how could I miss that. 😐 🌴 
Thanks!

jsocol:
Glad it works for you! Going to close this out. If you feel like there's a way to make things like this clearer in the docs, definitely feel free to open a docs PR or issue

PR #273, opened 2018-01-31
Fix typo in command to run linter

ndibari-sp:
In `run.sh`, the usage command describes the command to run the linter as `link`, but the switch/case block is looking for `lint`. This updates the usage function to print out the proper command to run the linter in the waffle-django project


clintonb (approved):

jsocol:
Thanks!

PR #277, opened 2018-03-04
Added support for i18n

clintonb:
The project now supports i18n with translations from Transifex.


clintonb:
FYI @KostyaEsmukov. 

I setup a Transifex project at https://www.transifex.com/django-waffle/django-waffle/.

KostyaEsmukov:
Wow, thanks for the prompt implementation! I've accepted the Transifex invitation and I'm going to add translations soon.

KostyaEsmukov (commented):
Could you please add `verbose_name` and `verbose_name_plural` to all models? Also having `verbose_name` on all model fields would be great.

Thanks!

waffle/models.py
```diff
@@ -120,41 +120,41 @@ class Flag(BaseModel):
...
         'Allow this flag to be set for a session for user testing.'))
-    superusers = models.BooleanField(default=True, help_text=(
+    superusers = models.BooleanField(default=True, help_text=_(
         'Flag always active for superusers?'))
-    staff = models.BooleanField(default=False, help_text=(
+    staff = models.BooleanField(default=False, help_text=_(
         'Flag always active for staff?'))
-    authenticated = models.BooleanField(default=False, help_text=(
+    authenticated = models.BooleanField(default=False, help_text=_(
         'Flag always active for authenticate users?'))
```
KostyaEsmukov: I think it should be `authenticated` - the trailing `d` is missing

waffle/models.py
```diff
@@ -119,49 +119,93 @@ class Flag(BaseModel):
...
-        settings.AUTH_USER_MODEL, blank=True,
-        help_text=_('Activate this flag for these users.')
+        settings.AUTH_USER_MODEL,
+        blank=True,
+        help_text=_('Activate this flag for these users.'),
+        verbose_name=_('Users'),
+    )
+    rollout = models.BooleanField(
+        default=False,
+        help_text=('Activate roll-out mode?'),
```
KostyaEsmukov: `_` is missing here. Looks like this is the only missing piece.

clintonb: Fixed.

KostyaEsmukov:
The last string is translated. Thank you for this work! 

This looks complete to me. But please don't merge yet: I want to manually browse the translated interface to ensure that everything is ok. I'll let you know when I'm done (hopefully I'll be able to do this today. Tomorrow at most)

clintonb:
Strings updated. No worries. I can't merge until @jsocol, or another user with write access, approves this pull request.

waffle/models.py
```diff
@@ -119,49 +118,93 @@ class Flag(BaseModel):
...
+        help_text=_('Flag always active for authenticated users?'),
+        verbose_name=_('Authenticated'),
+    )
+    languages = models.TextField(
+        blank=True,
+        default='',
+        help_text=_('Activate this flag for users with one of these languages (comma-separated list)'),
+        verbose_name=_('Languages'),
+    )
+    groups = models.ManyToManyField(Group, blank=True, help_text=_(
```
KostyaEsmukov: `verbose_name` is missing here

KostyaEsmukov (commented):
I've checked the result and it seems to look good to me. I've made some minor modifications at Transifex.

Also `MANIFEST.in` file should include the following line for the language files to be copied with the package:

```
recursive-include waffle/locale *.mo
```

jsocol:
> I can't merge until @jsocol, or another user with write access, approves this pull request.

I can take a look tonight

clintonb:
@jsocol let me know if you're still able to take a look today, or when you can get to it later this week.

waffle/locale/en_US/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,206 @@
+# SOME DESCRIPTIVE TITLE.
```
jsocol, line 1: Should these be filled out?

jsocol (approved):
one question but lgtm!

waffle/models.py
```diff
@@ -119,49 +118,97 @@ class Flag(BaseModel):
...
+        blank=True,
+        default='',
+        help_text=_('Activate this flag for users with one of these languages (comma-separated list)'),
+        verbose_name=_('Languages'),
+    )
+    groups = models.ManyToManyField(
+        Group,
+        blank=True,
+        help_text=_('Activate this flag for these user groups.'),
+        verbose_name=_('Group'),
```
KostyaEsmukov: Shouldn't it be a plural `Groups`, like the `Users` field below?

waffle/locale/en_US/LC_MESSAGES/django.po
clintonb: In all these years, I have never noticed that line. It doesn't show up in Transifex, and I see no method of modifying it with the `makemessages` command.

waffle/models.py
clintonb: Yes. Fixing...

clintonb: Pushed.

waffle/locale/en_US/LC_MESSAGES/django.po
KostyaEsmukov: Agreed, IMO it's OK to leave it as-is. 
Here's the same line from DRF, using Transifex as well: https://github.com/encode/django-rest-framework/blob/d2994e0596c3163ac970b29dad6a61485f938045/rest_framework/locale/en/LC_MESSAGES/django.po#L1

waffle/models.py
KostyaEsmukov: Translated 👍

KostyaEsmukov (approved):
LGTM, thanks again!

Issue #220, opened 2016-07-14
Stricter settings flip


mallyvai:
Django-waffle is awesome! Feature request - Looks like there are default values assigned for switches that aren't defined in the database yet. I'd love some settings to either/or:

1) Logging any switches/flags/etc that are attempted on read but undefined in the database 
2) A setting that fails the request if the setting isn't retrievable

It's easy for feature-flags to sort of get strewn about the code, and this sort of thing can help enforce discipline / rigor with their implementation.

jsocol:
Logging non-existent settings is a great idea! We can probably just log them to a `waffle` logger at a `WARNING` level and let folks configure/silence the logger if they want to.

koliber:
Logging nonexistent flag checks is a valuable improvement. The PR would be tin, seemingly uncontentious,  and easy to merge in. 

Is there anything that needs to be done before such a PR can be made?

bmoelans:
See https://github.com/django-waffle/django-waffle/pull/310 that solves logging issue, it is a change we already use as "fork" at our company

Main reasons:
* is more likely you add good description to it
* you can find in history who created that feature flipper and thus plausible can give you more details

PR #282, opened 2018-03-19
Fix issue #220 - Log non-existent settings as warning

charleswhchan:
How does this look?

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -2,6 +2,7 @@ from __future__ import unicode_literals
 
 import random
 from decimal import Decimal
+import logging
 
 from django.conf import settings
 from django.contrib.auth.models import Group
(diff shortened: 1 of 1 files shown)


waffle/models.py
```diff
@@ -13,6 +14,9 @@
 from waffle import managers
 from waffle.utils import get_setting, keyfmt, get_cache
 
+LOGGER_NAME = 'waffle'
+logger = logging.getLogger(LOGGER_NAME)
```
jsocol: Can this be a little simpler, just `logger = logging.getLogger('waffle')`? We're not re-using `LOGGER_NAME` anywhere and contextually it's obvious that it's the logger name.

jsocol (commented):
one small note but otherwise looks good, thanks!

charleswhchan:
@jsocol: PR updated. Please review again. 😃

jsocol (approved):

jsocol:
Thank you!

charleswhchan:
Thanks @jsocol 👍

PR #283, opened 2018-03-20
remove dot from admin actions

peterdemin:


clintonb (approved):
“Period” to be more specific

PR #285, opened 2018-03-28
Allow serializing/deserializing waffle models by natural key

bhomnick:
We find it useful when building fixtures between environments to use natural keys to avoid collisions with incremental integer PKs. This patch adds support for dumping/loading waffle models by name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,5 @@ dist
 .pydevproject
 .settings
 .tox/
+.DS_Store
 docs/_*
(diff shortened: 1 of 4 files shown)


clintonb (changes requested):
I'm not convinced this is necessary. You can already call `.get(name='some-string')`. Why do you need an alias for that call?

bhomnick:
@clintonb These are used by the `manage.py dumpdata --natural-primary` and `manage.py loaddata` commands to generate and load fixtures.

https://docs.djangoproject.com/en/2.0/topics/serialization/#topics-serialization-natural-keys

clintonb (commented):
Thanks or the explanation. Please add a couple tests to ensure this functionality is not inadvertently modified in the future.

clintonb (approved):

PR #292, opened 2018-04-26
Custom flag models

### LINKED ISSUE #181: Abstract base models

### LINKED ISSUE #286: Customizable Flag's groups model
The `auth.Group` model is used often as user's role. Would be good if was possible to customize, through Django settings, the model used in `Flag.groups`. This way it wouldn't be required to use user's role to define whether he access to a feature. In my case I'd like to have a custom `FlagsGroup` (developed in my project scope) instead.

Would like to hear from the official contributors if you think this would be a relevant feature in this lib. I'm available to make a PR really soon.

hugobessa:
**WAFFLE_FLAG_MODEL**

The model that will be used to keep track of flags. Defaults to `waffle.Flag` which allows user- and group-based flags. Can be swapped for a different Flag model that allows flagging based on other relationships or properties, such as an Organization a user belongs to. Analogous functionality to Django's extendable User models. Needs to be set at the start of a project. 

`Obs.:` Migrations do not support changing after the initial migration.

This PR creates two abstract models to help users create their own Flag model with existing functionality:  `AbstractBaseFlag` and `AbstractUserFlag`.

This is the same code written by @koliber [here](https://github.com/jsocol/django-waffle/pull/243), I only rebased it.

fixes #181, fixes #286

### DIFF
--- a/docs/starting/configuring.rst
+++ b/docs/starting/configuring.rst
@@ -11,6 +11,15 @@ behavior.
     The format for the cookies Waffle sets. Must contain ``%s``.
     Defaults to ``dwf_%s``.
 
+``WAFFLE_FLAG_MODEL``
+    The model that will be used to keep track of flags. Defaults to ``waffle.Flag``
+    which allows user- and group-based flags. Can be swapped for a different Flag model
+    that allows flagging based on other relationships or properties, such as an Organization 
+    a user belongs to.
+    Analogous functionality to Django's extendable User models.
+    Needs to be set at the start of a project. Migrations do not support changing after the
+    initial migration.
+
 ``WAFFLE_FLAG_DEFAULT``
     When a Flag is undefined in the database, Waffle considers it
     ``False``.  Set this to ``True`` to make Waffle consider undefined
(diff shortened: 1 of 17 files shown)


clintonb:
See https://github.com/jsocol/django-waffle/pull/291

PR #294, opened 2018-04-26
Waffle flag switch calls

### LINKED ISSUE #288: Make it easier to write automatically-removable flag branches
To avoid nesting code with if/else when changing logic whether a flag/switch is active or not. 

My proposal is to decorate helper functions with a `waffle_call` decorator and make the braching with a helper function. 

```python
@waffle_callable
def test_func_a(a, b, c):
    return (a, b, c)

@waffle_callable
def test_func_b(a, c):
    return (a, c)

my_return = waffle_switch_call(
    'my_switch', active_callable=test_func_a(1,2,3), inactive_callable=test_func_b(1, 3))

# or

my_return = waffle_flag_call(
    request, 'my_flag', active_callable=test_func_a(1, 2, 3), inactive_callable=test_func_b(1, 3))
```

I think this code style will force the users of the lib of creating more readable and modular code.

But the main objective of this architecture here is to make easier to implement automatic removal of flags in the code base using Abstract Syntax Tree tools like [astroid](https://github.com/PyCQA/astroid).

hugobessa:
Creates a way to avoid nesting code with if/else when changing logic whether flag is active or not. This force the users of the lib of creating more readable and modular code.

It also makes easier to implement automatic removal of flags in  code base using Abstract Syntax Tree tools like astroid.

In this PR users now have a decorator `@waffle_callable` for marking functions as they are only running based on a flag/switch state. And two helpers: one for flag `waffle_flag_call()` and `waffle_switch_call()` where the user pass the WaffleCallables and the required parameter to check the flag/switch state (request and flag name for flags and switch name for switches).

fixes #288

### DIFF
--- a/docs/usage/helpers.rst
+++ b/docs/usage/helpers.rst
@@ -1,0 +1,76 @@
+.. _helpers:
+
+=============================
+Helpers for flags based logic
+=============================
+
+Django Waffle provide some helpers and decorators to make easy to adapt
+your logic to be based on flags/switches status, and also make easy to 
+remove these flags/switched from code after the feature is accepted or
+rejected.   
+
+
+Waffle callables
+================
+
+The best way to make a feature flip is by modularizing your code, so
+the active and inactive states logic don't get mixed. This way is 
+easier to read the code and to remove the feature flag/switch after.
+
+Django Waffle provides a ``@waffle_callable`` decorator to mark
+functions to be called or not by the helpers depending on the 
+flag/switch state at the moment.
+
+Eg.::
+    
+    from waffle.decorators import waffle_callable
+
+    @waffle_callable
+    def my_great_function(a, b, c):
+        return [a, b, c]
+
+
+Flags
+=====
+
+There is a ``waffle_flag_call()`` helper to branch logic depending on 
+flag state. It receives the request, the flag name, the active state
+waffle_callable and the inactive waffle_callable (optional).
+
+It checks whether the flag is active and calls the proper 
+``waffle_callable``::
+
+    waffle_flag_call(
+        request, 'my-flag', my_great_function(1, 2, 3), 
+        my_inactve_function())
+
+If you want to check if the feature is inactive to run your 
+waffle_callable, you can add an exclamation point before your flag 
+name::
+
+    from waffle import waffle_flag_call
+
+    waffle_flag_call(
+        request, '!my-flag', my_inactve_function())
+
+Switches
+========
+
+The logic is the same as for flags. There's a ``waffle_switch_call()``
+helper to branch logic depending on the flag  state. It receives the 
+the flag name, the active state waffle_callable and the inactive 
+waffle_callable (optional).
+
+It checks whether the flag is active and calls the proper 
+``waffle_callable``::
+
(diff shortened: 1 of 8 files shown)


jsocol:
I don't love this API, honestly. It seems to add a lot of indirection and line length for not much value.

What's the need for `@waffle_callable` as a wrapper? Can someone execute these functions outside waffle?

hugobessa:
The `@waffle_callable` is just to save the arguments for a future execution (inside the helpers, depending on the flag/switch state). You could do this with a lambda function, but that would be harder to find when you're removing the flag of code.

What I'm trying to accomplish here is to make easier to remove the flag, which is an important part of the process of using feature flags. Having old flags is bad because they inactive state usually tends to break after some time and also because your code is harder to read with flags (specially old ones). Unfortunately, removing feature flags is a task that is usually postponed indefinitely. 

Well, this was my intention, but I'd really love to hear what's your opinion about this. Maybe we can think of other way to make flags removal more automatic in this discussion.

hugobessa:
@jsocol do you have any thoughts on this topic? @clintonb and @koliber, I'd love to hear from you too.

koliber:
I will try to find time to look at this this week.

koliber:
I agree that removing flag code is an important part of code maintenance. In the absence of this solution, people have been dealing with it in various ways:

1. not cleaning up feature flags that are no longer in use
2. removing unused feature flag code piecemeal, as they find it, during other coding activities
3. using their IDE's or code editor's find-and-replace functionality to find string literals and remove feature flags in one go

I am a proponent of keeping things simple. I think the proposed mechanism is complicated and does not add much benefit over option 3: find & replace. The cost of not removing features thoroughly is not high, so there is no pressing need for a comprehensive solution for unused feature flag removal.

Find & replace is most effective if features have unique and easy-to-find names. This is something that can be taught through Waffle documentation. There could be section on naming flags and removing ones that are no longer used. 

The proposed mechanism is not easy to understand. This will limit its use greatly. Good documentation will ease this.

I am trying to imagine a company with a code base with 20 active developers, both front and back end. One of the senior devs finds this solution and implements it in the code base. They have now gained the ability to remove features easily. However, they have introduces complexity in adding feature flags. I would venture that half the devs will have a hard time understanding why they need to write feature flag code this way. I don't think it would be a net positive for that organization to adopt this way of coding. The benefit does not outweigh the cost.

jsocol:
I just read a post (that of course I can't find again) about strategies for removing flags, and all of them were processes, not tech.

I think my concern here is very closely aligned with @koliber, with one thing in particular standing out: This is a tooling question. I support teams finding ways to automate removing flags, but I don't want to privilege/require a particular solution. There's not one standard "codemod" style tool (at least not that I'm aware of, but I admit I've been more in JS land for a while).

So I think code like this best lives in userland, along with the associated tooling to use it, for the time being. I'm going to close the issue and PR, but thank you for the suggestion and discussion!

hugobessa:
Thank you for the discussion, folks! I see your point, and totally agree this should be in "userland".  I'll work your documentation suggestions, @koliber, and make a PR in the future.

PR #297, opened 2018-07-04
Add Python3.7 to tests

adamkal:


tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 envlist =
-    py{27,34,35,36}-django111
-    py{34,35,36}-django20
+    py{27,34,35,36,37}-django111
```
clintonb: Django 1.11.x does not officially support Python 3.7.x.

clintonb (changes requested):
Remove the Django 1.11 change, and I'll approve.

tox.ini
adamkal: oh yeah! You are right. Thx :+1:

clintonb (approved):

Issue #296, opened 2018-05-31
Cached values may be stale after updating in a transaction


dtao:
On Bitbucket we use waffle flags/samples/switches in some very high-traffic scenarios, which can lead to race conditions like [this one](https://github.com/jsocol/django-waffle/pull/249).

Recently we uncovered another one: since BaseModel.save() [calls flush() immediately](https://github.com/jsocol/django-waffle/blob/v0.14.0/waffle/models.py#L98) after saving, there is the potential for a separate process to read the previous value and cache it before the transaction is committed, resulting in the stale value being cached indefinitely.

You might think this is an edge case but in fact [the Django admin site always wraps changes in transactions](https://github.com/django/django/blob/2.0.5/django/contrib/admin/options.py#L1449-L1450), so this applies to any application for which the primary mechanism for flipping flags and switches is the admin UI.

The most natural fix for this would be to change `BaseModel.save()` to something like:

```python
from django.db import transaction

def save(self, *args, **kwargs):
    self.modified = timezone.now()
    ret = super(BaseModel, self).save(*args, **kwargs)
    transaction.on_commit(self.flush)
    return ret
```

However, `transaction.on_commit` [only became available in Django 1.9](https://docs.djangoproject.com/en/1.9/topics/db/transactions/#performing-actions-after-commit), so this wouldn't work out of the box on previous versions of Django. If this were my library, I'd probably just add a runtime check to use `on_commit` if available, otherwise call `flush()` directly and add a note in the docs to library consumers: sorry, but there is a race condition unless you're on Django>=1.9.

The reason I am opening an _issue_ and not a pull request is actually something else entirely: the current test suite seems to use sqlite, but the best way I can come up with to demonstrate the problem is to use threads and it seems sqlite [doesn't play nicely with threads](https://code.djangoproject.com/ticket/12118)—or at least, it didn't work for me locally (the linked issue appears to have a solution, but this is about where I stopped digging).

In any case, here is a test that can be added if someone wants to put in the effort to make it work:

```python
import threading

from django.db import transaction
from django.test import TransactionTestCase

from waffle import switch_is_active
from waffle.models import Switch


class WaffleTestCase(TransactionTestCase):
    def test_update_switch_in_transaction(self):
        """Wait to invalidate the cache until after the current transaction."""

        switch_name = 'transaction-switch-name'
        switch = Switch.objects.create(name=switch_name, active=False)
        self.addCleanup(switch.delete)

[...]

PR #299, opened 2018-07-14
Add support for running tests with Postgres

dtao:
This adds support for setting the DATABASE_URL environment variable to run tests with database engines other than sqlite3 locally. It also configures the CI build to run the tests with Postgres.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,24 @@ python:
   - "3.6"
 
 matrix:
- include:
-  - python: 3.6
-    env: TOXENV=i18n
+  include:
+    - python: 3.6
+      env: TOXENV=i18n
 
 install:
- - pip install tox tox-travis
+  - pip install tox tox-travis
+
+# Some tests may require accessing the database from multiple threads
+# (see #296), which isn't supported by the sqlite3 engine. Run the tests with
+# Postgres in CI to allow full test coverage.
+services:
+  postgresql
+
+before_script:
+  - psql -c 'create database waffle_test;' -U postgres
+
+env:
+  - DATABASE_URL=postgres://postgres@localhost:5432/waffle_test
 
 script:
- - tox
+  - tox
--- a/test_settings.py
+++ b/test_settings.py
@@ -28,6 +28,16 @@ DATABASES = {
     }
 }
 
+if 'DATABASE_URL' in os.environ:
+    try:
+        import dj_database_url
+        import psycopg2
+        DATABASES['default'] = dj_database_url.config()
+    except ImportError:
+        raise ImportError('Using the DATABASE_URL variable requires '
+                          'dj-database-url and psycopg2. Try:\n\npip install '
+                          '-r travis.txt')
+
 INSTALLED_APPS = (
     'django.contrib.admin',
     'django.contrib.auth',
--- a/tox.ini
+++ b/tox.ini
@@ -8,6 +8,7 @@ deps =
     django111: Django>=1.11,<2.0
     django20: Django>=2.0,<2.1
     -rtravis.txt
+passenv = DATABASE_URL
 commands =
     ./run.sh test
 
--- a/travis.txt
+++ b/travis.txt
@@ -2,6 +2,8 @@
 # versions.
 mock==1.3.0
 Jinja2>=2.7.1
+dj-database-url==0.5.0
 django-jinja>=2.1,<3
+psycopg2>=2.7.5
 
 transifex-client


dtao:
Note that this is actually a precursor to the fix for #296. As discussed on that ticket, the test for the fix involves threads, which don't work with sqlite3; so this sets up the Travis build to run the tests with Postgres.

I am doing this as its own PR because these changes are independent of the fix for #296, and it didn't feel right to me to lump them all together in a single pull request.

.travis.yml
```diff
@@ -9,12 +9,21 @@ python:
   - "3.6"
 
 matrix:
- include:
-  - python: 3.6
-    env: TOXENV=i18n
+  include:
```
dtao, line 12: The indentation in this file was inconsistent. The Python versions were indented with two spaces, but everything else was indented with one space.

I opted to indent everything two spaces. Happy to revert these particular changes if they're objectionable for some reason.

test_settings.py
```diff
@@ -28,6 +29,16 @@
...
 
+if 'DATABASE_URL' in os.environ:
+    try:
+        import dj_database_url
+        import psycopg2
+        DATABASES['default'] = dj_database_url.config()
+    except ImportError:
+        print('Using the DATABASE_URL variable requires dj-database-url and '
+              'psycopg2. Try:\n\npip install -r travis.txt')
+        sys.exit(1)
```
dtao: My feeling is that it's better to error and exit in this scenario rather than falling back to a database backend that isn't what was specified for the environment.

tox.ini
```diff
@@ -8,6 +8,7 @@ deps =
     django111: Django>=1.11,<2.0
     django20: Django>=2.0,<2.1
     -rtravis.txt
+passenv = DATABASE_URL
```
dtao, line 11: By default, [tox only passes the `PATH` environment variable](https://tox.readthedocs.io/en/stable/example/basic.html#passing-down-environment-variables). This is necessary to get the value for `DATABASE_URL` defined in .travis.yml passed to tox.

travis.txt
```diff
@@ -2,6 +2,8 @@
 # versions.
 mock==1.3.0
 Jinja2>=2.7.1
+dj-database-url==0.5.0
```
dtao, line 5: This just provides a convenient way of parsing the `DATABASE_URL` environment variable. I felt it was fairly harmless to add a dependency here, since these are after all only the CI requirements.

travis.txt
```diff
@@ -2,6 +2,8 @@
...
 django-jinja>=2.1,<3
+psycopg2>=2.7.5
```
dtao, line 7: This is obviously to allow Postgres to be used as the DB engine.

dtao:
The CI build is failing, but I'm not sure why. As far as I know I followed [the instructions for setting up PostgreSQL in Travis](https://docs.travis-ci.com/user/database-setup/#PostgreSQL) correctly. The docs say:

> The default user for accessing the local PostgreSQL server is postgres with a blank password.

Yet the test failure seems to indicate:

```
django.db.utils.OperationalError: could not translate host name "postgres" to address: Name or service not known
```

So I don't know. I'm going to try explicitly setting the port number, but if that doesn't work I'm out of ideas. Maybe somebody else knows what magic incantation is required to get it working.

dtao:
Ah, I see my mistake now. I misread: the docs only say that the default _user_ is postgres; it seems the hostname to use to connect to the database server is just localhost.

Now the tests are passing :)

test_settings.py
clintonb: You could also either (a) let the exception bubble up or (b) re-raise after printing the error message. Either way, an exception at this point will halt execution.

clintonb (approved):
Add a comment to `.travis.yml` and/or the `README` to inform other developers that Travis tests use PostgreSQL, and explain why.

@jsocol any feedback? If not, I will merge in ~72 hours.

test_settings.py
jsocol: Agreed, better not to call `sys.exit()` and let it exit on its own by re-raising.

I don't love the implicit dependency on `psycopg2` just because someone has defined `DATABASE_URL` at all—it might be useful to run the tests against MariaDB or to use PyGreSQL. But I don't think it's worth blocking this on.

travis.txt
jsocol: I feel like I've implemented this myself with `urlparse` so many times. TIL!

jsocol (commented):
@clintonb Agree about `sys.exit()`, one other small but non-blocking comment. Thanks @dtao!

test_settings.py
dtao: I agree re: the dependency on psycopg2. I was taking a "Don't build it until you need it" approach, figuring that the next person who comes along and wants to run tests locally with a different database could submit their own pull request to make the logic to handle the `ImportError` more surgical. Perhaps that was selfish of me.

I'm happy to make the logic here a little more conservative, only requiring psycopg2 if the DB backend is actually Postgres.

jsocol: Nah, I think your approach is right, we can add if we need later

dtao:
OK, I've added a comment to .travis.yml as @clintonb suggested and changed the logic in test_settings.py to raise an `ImportError` rather than `sys.exit(1)`.

I squashed these commits because it's my understanding that is preferred for this library.

clintonb (approved):

PR #278, opened 2018-03-06
Added instructions for releasing a new version

clintonb:

PR #300, opened 2018-07-14
 Flush cache after the current transaction is committed (#296)

dtao:
This builds off the work in pull request #299 and fixes the issue raised in #296.

Note: review #299 before reviewing this.

### DIFF
--- a/waffle/tests/test_testutils.py
+++ b/waffle/tests/test_testutils.py
@@ -3,14 +3,14 @@ from __future__ import unicode_literals
 from decimal import Decimal
 
 from django.contrib.auth.models import AnonymousUser
-from django.test import TestCase, RequestFactory
+from django.test import TransactionTestCase, RequestFactory
 
 import waffle
 from waffle.models import Switch, Flag, Sample
 from waffle.testutils import override_switch, override_flag, override_sample
 
 
-class OverrideSwitchTests(TestCase):
+class OverrideSwitchTests(TransactionTestCase):
     def test_switch_existed_and_was_active(self):
         Switch.objects.create(name='foo', active=True)
 
@@ -94,7 +94,7 @@ def req():
     return r
 
 
-class OverrideFlagTests(TestCase):
+class OverrideFlagTests(TransactionTestCase):
     def test_flag_existed_and_was_active(self):
         Flag.objects.create(name='foo', everyone=True)
 
@@ -140,7 +140,7 @@ class OverrideFlagTests(TestCase):
         assert not Flag.objects.filter(name='foo').exists()
 
 
-class OverrideSampleTests(TestCase):
+class OverrideSampleTests(TransactionTestCase):
     def test_sample_existed_and_was_100(self):
         Sample.objects.create(name='foo', percent='100.0')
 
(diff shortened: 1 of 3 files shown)


waffle/tests/test_testutils.py
```diff
@@ -3,14 +3,14 @@
 from decimal import Decimal
 
 from django.contrib.auth.models import AnonymousUser
-from django.test import TestCase, RequestFactory
+from django.test import TransactionTestCase, RequestFactory
```
dtao, line 6: Django's `TestCase` class automatically wraps each test method in a transaction and rolls back the transaction after the method returns. Since updating models now doesn't flush the cache until after the transaction is committed, this breaks the logic in most of these tests. Inheriting from `TransactionTestCase` addresses the problem by removing Django's magical auto-rollback behavior, the trade-off being that the tests are a tad slower.

dtao:
I suppose you could technically merge this before merging #299; there would just effectively be no test coverage since the test added in this PR is skipped when running tests with sqlite.

waffle/tests/test_waffle.py
```diff
@@ -394,3 +397,47 @@ def test_read_from_write_db(self):
...
             # the cache and DB are in sync.
             assert waffle.sample_is_active(sample.name)
+
+
+class TransactionTests(TransactionTestCase):
+    @unittest.skipIf('sqlite3' in settings.DATABASES['default']['ENGINE'],
+                     'This test uses threads, which the sqlite3 DB engine '
+                     'does not support.')
+    def test_update_switch_in_transaction(self):
+        """Wait to invalidate the cache until after the current transaction."""
```
jsocol: I appreciate all the comments here, this is definitely not an obvious issue.

Would you mind adding a slightly longer description to the doc block (as a second paragraph) and include a link to #296?

Should we have similar tests for `Flag` and `Sample`? I know the behavior exists in `BaseModel` but that's an implementation detail, not a public API. If we needed to refactor it, especially in support of extensible models, it could change or break.

jsocol (approved):
Again, thanks @dtao! I think we should wait and merge #299 first then get this one in—I'd rather watch the new CI build on master succeed without this patch first.

waffle/tests/test_waffle.py
dtao: For sure, I can add another comment explicitly referencing #296.

I'm also happy to add equivalent tests for `Flag` and `Sample`.

dtao:
@jsocol Per your suggestions I've added an explanatory comment to the somewhat complex test method to cover this case and also updated the test code to cover flags, switches, and samples.

I squashed this work into one commit because I _think_ that's what you prefer, though I acknowledge that can make it hard to tell what changed since you last reviewed the code. If I've misunderstood, and for future reference I should not squash commits in pull requests that are under review, let me know.

dtao:
Hey @jsocol @clintonb just making sure: with #299 merged, is this PR waiting on me for anything right now? I'm happy to wait patiently; I just don't want to be neglectful if there's something I should be doing to get this PR merged that I've overlooked.

clintonb (approved):

dtao:
Is there any interest in releasing a new version with this fix?

We are just depending on a fork at the moment. It's always nice to return to the fold when we can :)

clintonb:
I can take the steps to start a release, but I can't actually merge the requisite changes without @jsocol. I've been trying to get merge rights, but I haven't received a response on #278.

I'd recommend either sticking with your fork, or using the latest commit on master for the time being, @dtao.

Issue #50, opened 2012-09-08
Be able to define switches and default states in settings


tobych:
Just like Gargoyle:

http://gargoyle.readthedocs.org/en/latest/install/index.html#default-switch-states

Although I think Gargoyle only adds the switches when they're used. So it's autocreation (see the other Issue), but with the autocreation using stuff from settings if available.

GARGOYLE_SWITCH_DEFAULTS = {
    'new_switch': {
      'is_active': True,
      'label': 'New Switch',
      'description': 'When you want the newness',
    },
    'funky_switch': {
      'is_active': False,
      'label': 'Funky Switch',
      'description': 'Controls the funkiness.',
    },
}

When I used Gargoyle I always committed an addition to this setting along with whatever switch-controlled feature I was adding, to provide documentation and so noone would need to add the switch manually. I'd actually prefer all the switches to be added at startup though (nonexistant ones).

Currently I'm getting away with ./manage.py loaddata waffle.json each time I deploy, but that's not ideal.

I'll probably end up doing this myself as we'll want this in production soon. Comments welcome though about what I might best do and how I might do it. Perhaps this should be done with WoLpH (Rick van Hattem) and #44.

I'm ONLY using switches, currently, by the way.

jsocol:
> I'm ONLY using switches, currently, by the way.

Anything that works for switches, and is applicable, will need to apply to all three types.

tobych:
So James, shall I go ahead and work on this? I'd start with just a write-up of what I'd do, perhaps with documentation, and a little about implementation. I'd check-in with Rick (see above), too.

Sveder:
(2 years later) Are you guys still in favor of this issue? I'm thinking about scalable ways to deploy waffle switches and sync our dev, staging and production environments with the same switch names. Auto creation seems too implicit to me. I can try and polish the hack I'm currently writing into a pull request if there is someone here to review and patch it.

jsocol:
I've never been 100% on board with doing this. I can't say for sure why I never answered @tobych two years ago (sorry, Toby!) but that's probably part of it.

One thing to keep in mind is I'm really unhappy with the core of waffle right now. It's one of my first Python projects and that shows, in mundane ways like code organization and bigger, harder to approach problems like extensibility. Every flag can't have two hundred attributes but that's the road we're on right now.

[...]

PR #301, opened 2018-09-18
Optionally create missing Flags, Samples and Switches

adamhaney:
Flags, Switches and Samples may be scattered throughout templates,
backend code or across mulitple frontend clients. This patch provides
a user with the option of having waffle create a missing Flag, Sample
or Switch when one is requested but does not exist.

The goal of this change is to save operators time by removing the need
to search through code for flag/switch/sample names and to reduce
communication requirements accross teams with of multiple developers
with multiple environments that might require the creation of a
flag/switch/sample in the database of every environment code is run
in.

# Behavior Changes
## Old Behavior
  1. Flag/Switch/Sample is requested
  2a. If found in db evaulate what the value should be for request
  2b. If missing return (FLAG|SWITCH|SAMPLE)_DEFAULT value
## New Behavior
  1. Flag/Switch/Sample is requested
  2a. If found in db evaulate what the value should be for request
  2b. Create Flag/Switch/Sample with
      value (FLAG|SWITCH|SAMPLE)_DEFAULT
  3. Return (FLAG|SWITCH|SAMPLE)_DEFAULT for this request, subsequent
  requests will be served using the value from the cache/db

# Further Discussion
  - #44 (https://github.com/jsocol/django-waffle/issues/44)
  - #50 (https://github.com/jsocol/django-waffle/issues/50)

# Future Work
This change intentionally narrowly scoped and as such does not provide
a mechanism to set a default value other
than (FLAG|SWITCH|SAMPLE)_DEFAULT for individual objects that are not
in the database although it could be expanded in the future to do so.


waffle/models.py
```diff
@@ -243,6 +243,19 @@ def _is_active_for_language(self, request):
...
     def is_active(self, request):
         if not self.pk:
+            if get_setting('CREATE_MISSING_FLAGS'):
+                try:
+                    with transaction.atomic():
+                        get_waffle_flag_model().objects.create(
+                            name=self.name,
+                            everyone=get_setting('FLAG_DEFAULT')
+                        )
+                except IntegrityError:
```
clintonb: What would cause these errors? If such an error occurs, I would rather see it bubble up that be squashed with an uninformative log line.

waffle/models.py
```diff
@@ -481,5 +508,23 @@ class Meta:
 
     def is_active(self):
         if not self.pk:
+            if get_setting('CREATE_MISSING_SAMPLE'):
+                default_percent = 0.0
+
+                if get_setting('SAMPLE_DEFAULT'):
+                    default_percent = 100.0
```
clintonb: You want to use the setting value. This can be combined with the line above: `get_setting('SAMPLE_DEFAULT', 0)`.

waffle/models.py
```diff
@@ -243,6 +243,19 @@ def _is_active_for_language(self, request):
 
     def is_active(self, request):
         if not self.pk:
+            if get_setting('CREATE_MISSING_FLAGS'):
+                try:
+                    with transaction.atomic():
```
clintonb: Why are you creating a transaction for all of these?

clintonb (changes requested):
I must admit that I am not a fan of implicitly creating data in this manner. You run the risk of creating a bunch of data that needs to later be cleaned up. Why not use a migration to create the flag/sample/switch?

waffle/models.py
adamhaney: I wasn't trying to just get a default value, or at least I don't think I was :)

In the defaults `SAMPLE_DEFAULT` is a boolean value, however there isn't a boolean option for a Sample, it's expressed as a percentage, so I'm saying if `SAMPLE_DEFAULT` is true then that implies that you want the sample value to be 100, if I just grabbed the default value I'm testing if it's defined.

I've never really used Samples in production so I may be misunderstanding their intended use.

waffle/models.py
adamhaney: The IntegrityError would be thrown if you attempt to create a flag name that already exists, which can happen if two processes both encounter a flag that hasn't been created and attempt to create it. I caught it because I my philosophy around this change set was very "first do no harm" since it's intended to be a convenience and the behavior of returning the FLAG_DEFAULT is unaffected by whether or not the flag could be created I thought it was best to catch it.

Still, you're the maintainer so if you don't like that explanation I can definitely change it.

waffle/models.py
adamhaney: In Postgres (not sure about other databases) if a cursor throws an IntegrityError then later interactions with the database will fail (even if you catch the exception) [1]. So I'm just trying to make sure that failure to create flags doesn't take down what would otherwise be a perfectly healthy request.

1. https://docs.djangoproject.com/en/2.1/topics/db/transactions/#controlling-transactions-explicitly

Although, upon further thought I could probably accomplish a similar thing by using `get_or_create` instead of just `create`.

waffle/models.py
clintonb-stripe: Got it. I've never used samples, either.

clintonb: Got it. I've never used samples, either.

adamhaney:
> I must admit that I am not a fan of implicitly creating data in this manner. You run the risk of creating a bunch of data that needs to later be cleaned up. Why not use a migration to create the flag/sample/switch?

I'm currently working on a project where the goal is to wrap every new feature in a flag, but sometimes things fall through the cracks and code ends up in the codebase that people don't realize is there, or that is turned off behind a flag that people forget the name of. Ironically because it's so easy to drop flags into templates, and python, and even javascript it's hard to get a top level view of every flag that's being evaluated. 

Also, in an effort to use descriptive flags that don't have name collisions we use names like, "jira_32_send_users_weekly_email" which is really easy to typo and then you end up with a subtly typo'd flag name in the admin that you *think* is turning a flag on or off but instead it's just quietly evaluating the default.

I completely acknowledge the risk of accidentally implicitly creating data, to maintain continuity with the way this library has been operating for the last couple of years I wouldn't suggest making this setting on by default. In general it would only impact users of the library that are explicitly asking to have data created for them.

adamhaney:
Also, if you're considering accepting this patch I did realize I need to write some regression tests before it's merged.

clintonb:
I would solve the problem with tests and/or a checklist to ensure flags are created as part of the development process. This would solve the issue of typos and forgetting flags for server-side code.

That said, I understand the desire to do things this way. Ping me once you have tests, and I'll review again.

adamhaney:
@clintonb tests are added, I covered testing creation with the setting enabled and positive and negative paths for defaults please let me know if there are any additional tests you'd like to see, and thanks for taking the time to review.

Also FWIW I deployed my fork to our production service this morning and haven't seen any problems (although that's obviously anecdotal).

waffle/models.py
clintonb: Log as a warning, but include the stack trace by passing `exc_info=True` in your call to `warning`. Do that in all of the new exception handlers

waffle/models.py
```diff
@@ -481,5 +508,23 @@ class Meta:
 
     def is_active(self):
         if not self.pk:
+            if get_setting('CREATE_MISSING_SAMPLES'):
+                default_percent = 0.0
+
+                if get_setting('SAMPLE_DEFAULT'):
```
clintonb: `default_percent = 100 if get_setting('SAMPLE_DEFAULT') else 0`

waffle/tests/test_waffle.py
```diff
@@ -334,6 +334,35 @@ def test_pluggable_model(self):
...
+    @override_settings(WAFFLE_CREATE_MISSING_FLAGS=True)
+    @override_settings(WAFFLE_FLAG_DEFAULT=False)
+    def test_flag_created_dynamically_default_false(self):
+        flag_model = waffle.get_waffle_flag_model()
+
+        assert flag_model.objects.count() ==  0
+        assert not waffle.flag_is_active(get(), 'my_dynamically_created_flag')
+        assert flag_model.objects.count() == 1
+
+        flag = flag_model.objects.all()[0]
```
clintonb: Combine into a single call: 
```python
FLAG_NAME = 'my_dynamically_created_flag'
flag = flag_model.objects.get(name=FLAG_NAME)```

File: waffle/tests/test_waffle.py [hunk 6]
```diff
@@ -334,6 +334,35 @@ def test_pluggable_model(self):
         self.assertEqual(b'off', response.content)
         assert 'dwf_myflag' not in response.cookies
 
+    @override_settings(WAFFLE_CREATE_MISSING_FLAGS=True)
+    @override_settings(WAFFLE_FLAG_DEFAULT=False)
+    def test_flag_created_dynamically_default_false(self):
+        flag_model = waffle.get_waffle_flag_model()
```
clintonb: Extract all of this into a helper method (e.g. `assert_flag_dynamically_created(expected_value)`).

waffle/tests/test_waffle.py
```diff
@@ -396,6 +425,28 @@ def test_read_from_write_db(self):
             # the cache and DB are in sync.
             assert waffle.switch_is_active(switch.name)
 
+    @override_settings(WAFFLE_CREATE_MISSING_SWITCHES=True)
+    @override_settings(WAFFLE_SWITCH_DEFAULT=False)
+    def test_switch_created_dynamically_false(self):
+        assert Switch.objects.count() ==  0
```
clintonb: Same notes as above: create a helper method, load the object by name.

waffle/tests/test_waffle.py
```diff
@@ -429,6 +480,31 @@ def test_read_from_write_db(self):
             # the cache and DB are in sync.
             assert waffle.sample_is_active(sample.name)
 
+    @override_settings(WAFFLE_CREATE_MISSING_SAMPLES=True)
+    @override_settings(WAFFLE_SAMPLE_DEFAULT=False)
+    def test_sample_created_dynamically_default_false(self):
+        assert Sample.objects.count() ==  0
```
clintonb: Same notes as above: create a helper method, load the object by name.

clintonb (changes requested):

adamhaney:
@clintonb I took forever to get back to this change set but I think I've addressed all your requested changes.

clintonb (approved):

clintonb:
@adamhaney would you add documentation, please?

adamhaney:
@clintonb documented all of the settings in the "Configuring Waffle" section and added a section about the optional auto creation behavior on each type's page. I'm Happy to make any revisions based on feedback.

docs/starting/configuring.rst
```diff
@@ -62,3 +62,19 @@ behavior.
 
 ``WAFFLE_CACHE_NAME``
     Which cache to use. Defaults to ``'default'``.
+
+``WAFFLE_CREATE_MISSING_FLAGS``
+    If waffle encounters a reference to a flag that is not in the database should it create it?
```
clintonb:
```suggestion
    If Waffle encounters a reference to a flag that is not in the database, should Waffle create the flag?
```

docs/starting/configuring.rst
```diff
@@ -62,3 +62,19 @@ behavior.
...
+    If true new flags are created and set to the value of ``WAFFLE_FLAG_DEFAULT``
+    Defaults to ``False``.
+
+``WAFFLE_CREATE_MISSING_SWITCHES``
+    If waffle encounters a reference to a switch that is not in the database should it create it?
```
clintonb:
```suggestion
    If Waffle encounters a reference to a switch that is not in the database, should Waffle create the sample?
```

docs/starting/configuring.rst
```diff
@@ -62,3 +62,19 @@ behavior.
...
+    If true new switchs are created and set to the value of ``WAFFLE_SWITCH_DEFAULT``
+    Defaults to ``False``.
+
+``WAFFLE_CREATE_MISSING_SAMPLES``
+    If waffle encounters a reference to a sample that is not in the database should it create it?
```
clintonb:
```suggestion
    If Waffle encounters a reference to a sample that is not in the database, should Waffle create the sample?
```

docs/types/sample.rst
```diff
@@ -52,3 +52,20 @@ Samples can be administered through the Django `admin site`_ or the
...
+sample returns the :ref:`WAFFLE_SAMPLE_DEFAULT <starting-configuring>`
+value but does not create a sample in the database. If you'd like
+waffle to create missing samples in the database whenever it
+encounters a missing sample you can set
+:ref:`WAFFLE_CREATE_MISSING_SSAMPLES <starting-configuring>` to
+``True``. If :ref:`WAFFLE_SAMPLE_DEFAULT` is ``True`` then the
+``Percent`` attribute of the sample will be created as 100.0 (so that
+when the sample is checked it always evalues to 
+``True``). Otherwise the value will be set to 0.0 so that the sample
+always evaluates to ``False``.`
```
clintonb:
```suggestion
always evaluates to ``False``.
```

docs/types/flag.rst
```diff
@@ -207,3 +207,19 @@ visitors" to "percent chance that the Flag will be activated per visit."
...
+
+When a flag is evaluated in code that is missing in the database the
+flag returns the :ref:`WAFFLE_FLAG_DEFAULT <starting-configuring>`
+value but does not create a flag in the database. If you'd like waffle
+to create missing flags in the database whenever it encounters a
+missing flag you can set :ref:`WAFFLE_CREATE_MISSING_FLAGS
+<starting-configuring>` to ``True``. Missing flags will be created in
+the database and the value of the ``Everyone`` flag attribute will be
+set to :ref:`WAFFLE_FLAG_DEFAULT <starting-configuring>` in the
+auto created database record.
```
clintonb:
```suggestion
auto-created database record.
```

docs/types/switch.rst
```diff
@@ -24,3 +24,18 @@ Switches can be administered through the Django `admin site`_ or the
...
+.. _types-switch-auto-create-missing:
+
+Auto Create Missing
+===================
+
+When a switch is evaluated in code that is missing in the database the
+switch returns the :ref:`WAFFLE_SWITCH_DEFAULT <starting-configuring>`
+value but does not create a switch in the database. If you'd like waffle
+to create missing switchs in the database whenever it encounters a
+missing switch you can set :ref:`WAFFLE_CREATE_MISSING_SWITCHS
```
clintonb:
```suggestion
missing switch you can set :ref:`WAFFLE_CREATE_MISSING_SWITCHES
```

docs/types/switch.rst
```diff
@@ -24,3 +24,18 @@ Switches can be administered through the Django `admin site`_ or the
...
+<starting-configuring>` to ``True``. Missing switchs will be created in
```
clintonb:
```suggestion
<starting-configuring>` to ``True``. Missing switches will be created in
```

docs/types/switch.rst
```diff
@@ -24,3 +24,18 @@ Switches can be administered through the Django `admin site`_ or the
...
+the database and the value of the ``Active`` switch attribute will be
+set to :ref:`WAFFLE_SWITCH_DEFAULT <starting-configuring>` in the
+auto created database record.
```
clintonb:
```suggestion
auto-created database record.
```

clintonb (changes requested):

guybowden:
I came here looking to see if anyone had created a set of simple migration commands (i.e. `manage.py makewaffleflagmigration SOME_FLAG_NAME`) to generate the instances - this looks like it'll negate the need for this. Good work guys.

clintonb (approved):

PR #310, opened 2018-11-07
Add log on missing flags samples and switches

bmoelans:
I saw https://github.com/django-waffle/django-waffle/issues/44 being released and definetly see advantages of this, but a while ago in our company we decided for another solution: log when it is missing.

Main reasons:
* is more likely you add good description to it
* you can find in history who created that feature flipper and thus plausible can give you more details


waffle/.flake8
```diff
@@ -0,0 +1,9 @@
+[flake8]
+; Use a line length of 120 instead of pep8's default 79
+max-line-length = 120
+; Files not checked:
+;  - .*: don't check 'hidden' directories
+;  - migrations: most of these are autogenerated and don't need a check
+;  - manage: these are autogenerated and don't need a check
+;  - docs: contains autogenerated code that doesn't need a check
+exclude = .?*,*migrations,docs
```
clintonb: Add a newline.

clintonb (approved):

waffle/.flake8
bmoelans: Newline added

Issue #311, opened 2018-11-16
Specify user list for flag using CLI


acraftydev:
I'm trying to administer a flag using the cli but currently there is no option to specify a list of users only list of groups. I'm looking at waffle/management/commands/waffle_flag.py and can't see any parsing for a user switch?

I'm happy to try and patch it but wanted to make my understanding was correct.

clintonb:
@crafty78, you are correct: there is no support for passing a list of users to that command. Is there a reason you prefer to pass in users rather than use a group? I'm open to updating the command, but I want to better understand your use case.

acraftydev:
Convenience, we often trial features quickly with a couple of individual users before large groups, so being be able to toggle for user rather than having to manage a group is quickest. Also i'm not aware of a built-in django auth admin cli for managing groups (ie creating/adding/removing) and we do our admin via the cli.

clintonb:
Post a PR, and I'll review this weekend.

PR #312, opened 2018-11-20
add user as command line option for admin command

acraftydev:
#311 add option to admin cli command to allow users to be added to feature flag.


clintonb (changes requested):
Please add a test for the case where `append` is `False`.

clintonb (approved):

PR #315, opened 2018-12-13
WIP/RFC: add simple support for multiple checks in template tags

costela:
This adds support for checking multiple flags/switches/samples in a single template tag. E.g. the following block would be shown when either of the two flags `featureA` or `featureB` is on:
```
{% flagany "featureA" "featureB" %}
...
{% endflagany %}
```

AFAICT, this currently cannot be so easily achieved. The common suggestion in this case seems to be flag nesting, but this is cumbersome because the `else` block would need to be duplicated.

This enables the use case where multiple teams are working on separate feature flags that share some common changes, which - for instance - should be enabled if any of the dependent features flags is enabled.

I'm not sure the current API (introducing separate template tags for each logical operator) is the best solution. It's a compromise to avoid adding full-fledged logical expression parsing to the template tags, which seemed like the easiest approach, but I'm open for other suggestions, of course.

Also, I still didn't add any tests for this :confused:

### DIFF
--- a/waffle/templatetags/waffle_tags.py
+++ b/waffle/templatetags/waffle_tags.py
@@ -10,16 +10,22 @@ from waffle.views import _generate_waffle_js
 register = template.Library()
 
 
+# fake logic operator for backward compatibility with single-argument template tags
+def _first(l):
+    return bool(next(iter(l)))
+
+
 class WaffleNode(template.Node):
     child_nodelists = ('nodelist_true', 'nodelist_false')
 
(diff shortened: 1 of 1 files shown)


costela:
hi @clintonb!
just for future reference it would be great if we could have some context about the PR closure :pray: 

Has this feature been obsoleted by something else in the meantime? Is there simply no interest in it?

clintonb:
The pull request has been WIP for nearly three years without any progress. I’m happy to merge once tests are added. Otherwise, there doesn’t seem to be any interest.

PR #316, opened 2018-12-13
Improve admin site for users and groups.

### LINKED ISSUE #313: Hard to see which users and groups a flag is activated for
The admin panel only shows the primary key (integer) of the users and groups the flag is activated for. It takes a while to determine who they are.

It would be great if the user and group names could be displayed.

PetterS:
- Use the regular list widget for groups (there are likely not
  that many, so performance should not be an issue).
- Keep the users as a raw_id field, but display the selected users
  after the input box in a parenthesized list.

Fixes #313.

Before
===
before

After
===
after

### DIFF
--- a/waffle/admin.py
+++ b/waffle/admin.py
@@ -1,6 +1,8 @@
 from __future__ import unicode_literals
 
 from django.contrib import admin
+from django.contrib.admin.widgets import ManyToManyRawIdWidget
+from django.utils.html import escape
 from django.utils.translation import ugettext_lazy as _
 
 from waffle.models import Flag, Sample, Switch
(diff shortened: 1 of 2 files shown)


PetterS:
Will likely have to do some work to support e.g. Python 2.7 with this patch. Unicode user names, specifically.

clintonb:
@PetterS would you post some before/after screenshots, please?

PetterS:
@clintonb sure, I updated the description.

waffle/tests/test_admin.py
```diff
@@ -0,0 +1,33 @@
+from __future__ import unicode_literals
+from unittest import mock
```
clintonb:
```python
try:
    import mock
except ImportError:
    import unittest.mock as mock
```

clintonb (changes requested):
Good to go once the import is fixed.

waffle/tests/test_admin.py
PetterS: Done.

clintonb (approved):

PetterS:
Let me see if I can create a python2 environment and fix the errors.

waffle/admin.py
```diff
@@ -41,14 +43,44 @@ def delete_individually(ma, request, qs):
...
 
+    def formfield_for_dbfield(self, db_field, **kwargs):
+        if db_field.name == 'users':
+            kwargs.pop('request', None)
+            kwargs['widget'] = \
+                InformativeManyToManyRawIdWidget(db_field.remote_field,
+                                                 self.admin_site,
+                                                 using=kwargs.get("using"))
+            return db_field.formfield(**kwargs)
+        return super().formfield_for_dbfield(db_field, **kwargs)
```
cleder: `return super(FlagAdmin, self).formfield_for_dbfield(db_field, **kwargs)` for python 2 compatibility

PetterS: Now I ran the tests with both 2.7 and 3.6.

clintonb (approved):

PetterS:
Anything left to do here? This should be based on the latest commit to master.

PetterS:
Thanks a lot @clintonb! Really helpful for us!

PR #318, opened 2019-01-02
Make `override_switch` usable with coroutines

godlark:
I've encountered a situation when I needed a wrapper for coroutines. I had to use it in async tests. That's why I create a PR to add the functionality to main repo.

A wrapper for coroutines has to be different than for normal functions.

### DIFF
--- a/waffle/tests/test_testutils.py
+++ b/waffle/tests/test_testutils.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 
+import asyncio
 from decimal import Decimal
 
 from django.contrib.auth.models import AnonymousUser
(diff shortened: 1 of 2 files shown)


clintonb (changes requested):
This needs to retain backward-compatibility with Python 2.7. Please address the build failures, and squash to a single commit. 

I’m not sure why this is needed. Why are your test methods async, as opposed to the code being tested? Can you share an example of how someone might use this?

schinckel:
I don't believe this solution worked anyway - it does not address the fact that the decorators need to be explicitly changed to support async. (ie, you can't just do a .get() in the async function call).

PR #323, opened 2019-01-11
Fix reported version to 0.15.2

PetterS:
The version is displayed in e.g. the Django debug panel and is reported incorrectly for 0.15.1, creating slight confusion about the environment.

I set it to 0.15.2 so it will be correct after next release?


clintonb (approved):

PetterS:
Merged with master.

PR #326, opened 2019-01-29
Flush cache after adding/removing users and groups

dtao:
Currently, the cached value for a flag is not invalidated when users
or groups are added or removed. This fixes that by registering receivers
for the corresponding m2m_changed signals.

A new signals module is introduced for this purpose, and imported in the
ready method of a new AppConfig class. This is the recommended way of
defining signals in all of the currently supported versions of Django:

https://docs.djangoproject.com/en/1.11/topics/signals/

### DIFF
--- a/waffle/__init__.py
+++ b/waffle/__init__.py
@@ -7,6 +7,7 @@ from django.apps import apps as django_apps
 
 VERSION = (0, 13, 0)
 __version__ = '.'.join(map(str, VERSION))
+default_app_config = 'waffle.apps.WaffleConfig'
 
 
 def flag_is_active(request, flag_name):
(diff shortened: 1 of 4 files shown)


waffle/__init__.py
```diff
@@ -7,6 +7,7 @@
 
 VERSION = (0, 13, 0)
```
dtao, line 8: This is also something I noticed. You might want to fix this! 😬

dtao: Ah, I see someone else [beat me to it](https://github.com/django-waffle/django-waffle/pull/323)!

clintonb (approved):
Thanks!

PR #331, opened 2019-02-23
Allow for inheritance of decorated classes

slafs:
When using `waffle.testutils.override_*` decorators on a class
one would expect the change to be visible on child classes too
(i.e. inheritance would work as normal).

This change drops the rather hacky implementation of
`testutils._overrider.for_class` method in favor of Django's
`django.test.utils.TestContextDecorator`
(the same one that's used as a base for `override_settings` utility).

The most notable change here is that `TestContextDecorator`
hooks up with unittest's `setUp`/`tearDown` mechanism
instead of decorating each test method in a decorated class.

**Backwards incompatible change (!!!):**
Only subclasses of `unittest.case.TestCase` can be decorated
with `testutils.override_*` decorators now.
This means for example that mixin-like test classes (inheriting from
`object` only) can't be decorated like that anymore.

### DIFF
--- a/waffle/tests/test_testutils.py
+++ b/waffle/tests/test_testutils.py
@@ -344,18 +344,30 @@ class OverrideSampleOnClassTransactionTestCase(OverrideSampleOnClassTestsMixin,
 
 
 class InheritanceOverrideSwitchOnClassTests(OverrideSwitchOnClassTestCase):
+    """
+    Extend ``OverrideSwitchOnClassTestCase``
+    and make sure ``override_switch`` change still works.
+    """
 
     def test_child_undecorated_method_is_set_properly_for_switch(self):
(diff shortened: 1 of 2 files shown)


slafs:
Heh, 3 minutes after opening this PR a substantial change of `test_testutils.py` was merged from #329 😅 - I'll try to update accordingly.

slafs:
This is now ready for review! :)

clintonb-stripe (commented):
Wrong account!

clintonb (commented):
Would you add a comment to your `InhertanceOverride` test classes explaining what you're doing in the tests? It took me a few minutes to realize exactly what you're doing.

slafs:
Good point. Docstrings added. I'm open for hints on adjusting the wording there.

clintonb (approved):

PR #333, opened 2019-04-02
Add support for Django 2.1

dbowring:
Add Django 2.1 to `tox.ini`. Tests pass without changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,9 @@ install:
 services:
   postgresql
 
+addons:
+  postgresql: "9.4"
+
 before_script:
   - psql -c 'create database waffle_test;' -U postgres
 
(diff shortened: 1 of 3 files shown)


dbowring:
These failures are caused by using postgres<9.4. I'm not familiar with travis-ci, but upgrading postgres should fix those errors.

clintonb:
Yep. I just found https://groups.google.com/forum/#!topic/django-users/behDaxyzVE8.

clintonb:
Try adding this to `.travsi.yml`:

```yaml
addons:
  postgresql: 9.4
```

https://docs.travis-ci.com/user/database-setup/#using-a-different-postgresql-version

clintonb:
Would you also update `setup.py`, please?

https://github.com/django-waffle/django-waffle/blob/master/setup.py#L25

clintonb (approved):

PR #346, opened 2019-09-22
Improve performance of `is_active_for_user`.

jmduke:
Howdy! Thanks for a great package.  I have a small change for improving `is_active_for_user` query performance when checking flag enrollment for a flag without groups; namely, skipping the user.groups call (which generally results in a DB call.)

This should be a no-op; `set().intersection(<anything>) == set()`, so this should just be a easy little way to short-circuit.

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -377,9 +377,10 @@ class AbstractUserFlag(AbstractBaseFlag):
 
         if hasattr(user, 'groups'):
             group_ids = self._get_group_ids()
-            user_groups = set(user.groups.all().values_list('pk', flat=True))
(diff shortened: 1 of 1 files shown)


jmduke:
Here's an example before and after of the change. In the context processor for this view, I'm grabbing a list of views with `Flag.get_all()` and iterating over them. Think something like:

```
{flag.name: flag.is_active_for_user(obj.user) for flag in Flag.get_all()}
```

Before:

Screen Shot 2019-09-22 at 4 51 44 PM

After: 

Screen Shot 2019-09-22 at 4 52 05 PM

clintonb (approved):

PR #351, opened 2019-10-23
Add permissions for admin actions.

PetterS:
https://docs.djangoproject.com/en/2.2/ref/contrib/admin/actions/#setting-permissions-for-actions

Very important since otherwise users who only has view access can still modify flags.


clintonb (commented):
@PetterS is there an easy way to test these changes?

PetterS:
Yes, there is a public method on ModelAdmin `get_actions` that we can use. Hold on.

clintonb (approved):

PR #357, opened 2019-12-05
Add support for Django 3.0; drop support for ≤ 2.1

### LINKED ISSUE #354: django-waffle is not compatible with Django 3.0
Django 3.0 drops support for Python 2, causing issues like this:

```bash
  File "/usr/local/lib/python3.7/site-packages/waffle/models.py", line 11, in <module>
    from django.utils.encoding import python_2_unicode_compatible
ImportError: cannot import name 'python_2_unicode_compatible' from 'django.utils.encoding'
```

There's a full list of these python 2 compatibility APIs in the [Django 3.0 release notes](https://docs.djangoproject.com/en/3.0/releases/3.0/#removed-private-python-2-compatibility-apis).

PetterS:
Fixes #354.

If this is a better option than #355, we can do this.

All Django versions older than 2.2 are unsupported either at the start of 2020 or after the first quarter:
https://www.djangoproject.com/download/


tox.ini
```diff
@@ -1,16 +1,12 @@
 [tox]
 envlist =
-    py{27,34,35,36}-django111
-    py{34,35,36,37}-django20
-    py{35,36,37}-django21
     py{35,36,37}-django22
+    py{36,37}-django30
```
knyghty: Django 3.0 also supports Python 3.8

PetterS:
Looks like we have to wait for https://github.com/niwinz/django-jinja/pull/237

CHANGES
```diff
@@ -2,6 +2,11 @@
 Waffle Changelog
 ================
 
+v0.18.0
```
clintonb, line 5: `v0.19.0`

clintonb (commented):
We'll proceed with https://github.com/django-waffle/django-waffle/pull/355, once django-jinja is updated, to add Django 3.0 support. We can merge this shortly thereafter.

theunraveler:
Heads up that the PR should probably also replace all `ugettext` functions with `gettext` as noted [here](https://docs.djangoproject.com/en/3.0/releases/3.0/#id3). I've forked the repo and made the changes for reference, but didn't want to start another PR while this one was pending: https://github.com/theunraveler/django-waffle/commit/c9be929b4c5ef74779845b6574bd596afef75725.

clintonb:
Django 2.0 and 2.1 were dropped as part of #360. Thanks for kicking this off @PetterS!

PR #358, opened 2019-12-20
Made optional to test jinja2 due to django-jinja being unmaintained

NyanKiyoshi:
This change makes `django-jinja` as an optional feature thus can be skipped in tests.
All compatible django versions are still tested against jinja.

In #355, we can add `TOXENV='py{36,37,38}-django30-jinja'` to the test matrix and mark it as an allowed failure. When adding django 3.0, we will have `py{36,37,38}-django30` in the tox file to skip the jinja tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ python:
   - "3.5"
   - "3.6"
   - "3.7"
+  - "3.8"
 
 matrix:
   include:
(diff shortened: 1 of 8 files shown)


clintonb (approved):

PetterS:
Thanks!

PR #362, opened 2020-01-09
Remove note about dropping support for Django 2.2

nikolas:
Django 2.2 is still in LTS and is supported until April 2022. See: https://www.djangoproject.com/download/
So, django-waffle shouldn't drop support for it before then.

django-waffle is still being tested on Django 1.11 and Django 2.2. https://github.com/django-waffle/django-waffle/blob/master/.travis.yml#L9

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@ Waffle Changelog
 
 v0.19.0
 =======
-- Dropped support for Django 2.1, and Python 3.4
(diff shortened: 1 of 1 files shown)


clintonb (changes requested):
The note should read 2.0 and 2.1.

clintonb (approved):

PR #364, opened 2020-01-10
Use gettext_lazy instead of ugettext_lazy.

PetterS:
Removes a ton of warnings in Django 3.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,10 @@
 Waffle Changelog
 ================
 
+v0.19.1
+=======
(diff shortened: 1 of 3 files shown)


clintonb (commented):
Add a change log entry for 0.19.1 and ping me when you’re ready to merge.

PetterS:
@clintonb done!

clintonb (approved):

PetterS:
@clintonb would it be possible to have a new release so we can use this?

PR #366, opened 2020-01-16
Add category field to Switch and Flag models

### LINKED ISSUE #365: Feature: add category field to flag / switch models
We use waffle for multiple reasons - to hide work in progress, to release new features, to give beta access to other features, and to handle different customer "plans". As a result, we have a lot of flags in play, and they have become hard to manage.

It would be really helpful to have a `Flag.category` (`label`, `tag`, ...) field to be able to manage this.

* Add CharField to the Flag model (free text)
* Add field to admin list view
* Add field to admin list filters

hugorodgerbrown:
Fixes #365 - adding a category to the core models to make it easier to manage flags / switches if you are using them for multiple purposes. (I opened the PR without waiting for any approval on the suggestion as it was v. quick to implement.)

The motivation behind the issue is to have a simple mechanism for grouping flags / switches so that they can managed effectively when using them for different purposes - e.g. feature rollout, hiding WIP, customer segments, etc. We currently do this using flag prefixes - e.g. `WIP_new_feature_foo`, which becomes hard to manage after a while.

---

This PR adds a "Category" field to the `Flag` and `Switch` models, and exposes the category in the admin list view (in the list, and also as a list filter).

New Flag admin view:

Screenshot 2020-01-16 at 15 56 37

Admin list view:

Screenshot 2020-01-16 at 15 56 54

Admin list filter:

Screenshot 2020-01-16 at 15 57 07

---

NB I have also added a couple of extra django-admin commands to the `run.sh` script, as it helped me to test my changes, and I figured it may be useful to others. Happy to delete if inappropriate.


hugorodgerbrown:
Closing this as per @clintonb's comment on the source issue.

PR #372, opened 2020-06-02
Document `WAFFLE_TEST_COOKIE`

rik:


docs/starting/configuring.rst
```diff
@@ -9,7 +9,7 @@ behavior.
 
 ``WAFFLE_COOKIE``
     The format for the cookies Waffle sets. Must contain ``%s``.
-    Defaults to ``dwf_%s``.
+    Defaults to ``dwft_%s``.
```
clintonb, line 16: There are two formats, one of which is for testing. The existing format is correct. The test cookie format needs to be documented.

https://github.com/django-waffle/django-waffle/blob/06f98e2d5deec3cfd9a440b4cca868aefb38b11e/waffle/defaults.py#L4-L5

clintonb (changes requested):

docs/starting/configuring.rst
rik: Duh… Sorry for that, better commit coming.

clintonb (approved):

PR #378, opened 2020-08-18
Add support for Django 3.1+ and drop Django 1.11, 2.0

### LINKED ISSUE #376: AbstractBaseFlag uses deprecated NullBooleanField
[AbstractBaseFlag.everyone](https://github.com/django-waffle/django-waffle/blob/a8fc0f2ba636ab4520ae28d6dcb9d0d38f0b5905/waffle/models.py#L131) is a `NullBooleanField`. Django docs has been recommending to use `BooleanField(null=True)` and Django 3.1 pushed this even further making `NullBooleanField` deprecated.

hugorodgerbrown:
Fixes #376 

---

Django 3.1 introduces a deprecation warning for the `NullBooleanField`, and recommends switching to a `BooleanField` with `null=True` set. This field only supports `null` arguments from 2.1 on (https://docs.djangoproject.com/en/2.1/ref/models/fields/#booleanfield), which means it is not possible to support <2.1 and >3.1 together.

This PR drops 1.11, 2.0, 2.1 support. The package classifiers have been updated to reflect this.


CHANGES
```diff
@@ -2,6 +2,11 @@
 Waffle Changelog
 ================
 
+v1.1.0
```
clintonb: 2.0.0. We follow semantic versioning (https://semver.org/). Dropping support for a Django version is a major change.

clintonb (changes requested):
Thanks for the updates.

CHANGES
hugorodgerbrown: Do you want me to change the version in `setup.py` - or will you do that when you merge / release?

hugorodgerbrown: CHANGELOG has been updated.

clintonb (approved):

PetterS:
@clintonb do you think you could make a new release with this PR? Cheers!

PR #379, opened 2020-08-18
Cache auto-created missing flags

### LINKED ISSUE #343: Unnecessary database hits when CREATE_MISSING_SWITCHES is True and the switch does not exist
## What do I do

I'm calling repeatedly `switch_is_active` without adding the switch to DB. I also have `WAFFLE_CREATE_MISSING_SWITCHES` set to `True`

## What do I expect

I expect that even though I call `switch_is_active` n times, it hits the database only one time.

## What actually happens

`switch_is_active` hits the database every time I call it. 

## Why it happens

Here's the `switch_is_active` method body:

```
def switch_is_active(switch_name):
    from .models import Switch

    switch = Switch.get(sample_name)
    return switch.is_active()
```

If the switch does not exist, method `Switch.get` puts `CACHE_EMPTY` into the cache:

```
try:
   obj = cls.get_from_db(name)
except cls.DoesNotExist:
    cache.add(cache_key, CACHE_EMPTY)
    return cls(name=name)
``` 

And when cache key has the value `CACHE_EMPTY`, it returns an object without primary key:

```
if cached == CACHE_EMPTY:
    return cls(name=name)
``` 

And this is why `is_active` makes superfluous database query:

```
if not self.pk:
    # ...
    if get_setting('CREATE_MISSING_SWITCHES'): 
         Switch.objects.get_or_create(
             name=self.name,
             defaults={
                    'active': get_setting('SWITCH_DEFAULT')
             }
          )
```

## What do I propose

I propose caching the switch value upon calling `get_or_create`.

## Related

- https://github.com/django-waffle/django-waffle/issues/34

---

Let me know if you want me to fix it.

simonlindholm:
Fixes #343.

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -245,12 +245,14 @@ class AbstractBaseFlag(BaseModel):
             if log_level:
                 logger.log(log_level, 'Flag %s not found', self.name)
             if get_setting('CREATE_MISSING_FLAGS'):
-                get_waffle_flag_model().objects.get_or_create(
+                flag, _created = get_waffle_flag_model().objects.get_or_create(
                     name=self.name,
                     defaults={
                         'everyone': get_setting('FLAG_DEFAULT')
                     }
                 )
+                cache = get_cache()
+                cache.add(self._cache_key(self.name), flag)
 
             return get_setting('FLAG_DEFAULT')
 
@@ -444,12 +446,14 @@ class Switch(BaseModel):
             if log_level:
                 logger.log(log_level, 'Switch %s not found', self.name)
             if get_setting('CREATE_MISSING_SWITCHES'):
-                Switch.objects.get_or_create(
+                switch, _created = Switch.objects.get_or_create(
                     name=self.name,
                     defaults={
                         'active': get_setting('SWITCH_DEFAULT')
                     }
                 )
+                cache = get_cache()
+                cache.add(self._cache_key(self.name), switch)
 
             return get_setting('SWITCH_DEFAULT')
 
@@ -512,12 +516,14 @@ class Sample(BaseModel):
 
                 default_percent = 100 if get_setting('SAMPLE_DEFAULT') else 0
 
-                Sample.objects.get_or_create(
+                sample, _created = Sample.objects.get_or_create(
                     name=self.name,
                     defaults={
                         'percent': default_percent
                     }
                 )
+                cache = get_cache()
+                cache.add(self._cache_key(self.name), sample)
 
             return get_setting('SAMPLE_DEFAULT')
         return Decimal(str(random.uniform(0, 100))) <= self.percent
--- a/waffle/tests/test_waffle.py
+++ b/waffle/tests/test_waffle.py
@@ -51,7 +51,10 @@ class WaffleTests(TestCase):
         flag = flag_model.objects.get(name=FLAG_NAME)
         assert flag.name == FLAG_NAME
 
-        return flag
+        # We assert no queries are made to ensure flags created when the
+        # `CREATE_MISSING_FLAGS` setting is active are properly cached.
+        with self.assertNumQueries(0):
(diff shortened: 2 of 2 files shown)


waffle/models.py
```diff
@@ -244,12 +244,14 @@ def is_active(self, request):
             if log_level:
                 logger.log(log_level, 'Flag %s not found', self.name)
             if get_setting('CREATE_MISSING_FLAGS'):
-                get_waffle_flag_model().objects.get_or_create(
+                new_obj = get_waffle_flag_model().objects.get_or_create(
```
clintonb: This is a bit more standard in Python when accessing multiple returned objects than an index selector. Additionally, the variable name is more descriptive.

`flag, _created = get_waffle_flag_model().objects.get_or_create(`

waffle/models.py
```diff
@@ -244,12 +244,14 @@ def is_active(self, request):
...
                     name=self.name,
                     defaults={
                         'everyone': get_setting('FLAG_DEFAULT')
                     }
-                )
+                )[0]
```
clintonb: Ditch the array selector.

waffle/models.py
```diff
@@ -244,12 +244,14 @@ def is_active(self, request):
...
+                cache = get_cache()
+                cache.add(self._cache_key(self.name), new_obj)
```
clintonb: s/new_obj/flag

waffle/models.py
```diff
@@ -443,12 +445,14 @@ def is_active(self):
             if log_level:
                 logger.log(log_level, 'Switch %s not found', self.name)
             if get_setting('CREATE_MISSING_SWITCHES'):
-                Switch.objects.get_or_create(
+                new_obj = Switch.objects.get_or_create(
```
clintonb: Same as above, but change the variable name to `switch`.

waffle/models.py
```diff
@@ -443,12 +445,14 @@ def is_active(self):
...
                     name=self.name,
                     defaults={
                         'active': get_setting('SWITCH_DEFAULT')
                     }
-                )
+                )[0]
```
clintonb: Ditch the index selector.

waffle/models.py
```diff
@@ -443,12 +445,14 @@ def is_active(self):
...
+                cache = get_cache()
+                cache.add(self._cache_key(self.name), new_obj)
```
clintonb: s/new_obj/switch

waffle/models.py
```diff
@@ -511,12 +515,14 @@ def is_active(self):
 
                 default_percent = 100 if get_setting('SAMPLE_DEFAULT') else 0
 
-                Sample.objects.get_or_create(
+                new_obj = Sample.objects.get_or_create(
```
clintonb: Same as above...

waffle/tests/test_waffle.py
```diff
@@ -51,7 +51,8 @@ def assert_flag_dynamically_created_with_value(self, expected_value):
         flag = flag_model.objects.get(name=FLAG_NAME)
         assert flag.name == FLAG_NAME
 
-        return flag
+        with self.assertNumQueries(0):
```
clintonb, line 56: Add a comment above this context, explaining why it's here. I have a suspicion this may come back to bite us in the future. Something like this should be fine...

```
# We assert no queries are made to ensure flags created when the `CREATE_MISSING_FLAGS` setting is active  are properly cached.
```

Do the same for the other test updates, remembering to swap the correct component type (e.g., flag, sample, switch).

clintonb (changes requested):
Thanks for the changes. I'm happy to merge once the changes are made. Note that I may be a bit slower to respond as I am on parental leave.

simonlindholm:
I've made the requested changes.

clintonb (approved):

PR #384, opened 2020-12-14
Add Power Support ppc64le

nagesh4193:
Added power support for the travis.yml file with ppc64le. This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,23 @@ matrix:
       - { python: "3.8", env: TOXENV=py38-django31 }
 
       - { python: "3.7", env: TOXENV=i18n }
+      
+      # Adding power support architecture
+      - { arch: "ppc64le", python: "3.5", env: TOXENV=py35-django22, addons: {postgresql: "10"} }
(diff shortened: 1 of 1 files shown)


clintonb:
@nagesh4193 can you say more, please? Are you stating that django-waffle is part of an Ubuntu distribution? If so, for what purpose?

nagesh4193:
@clintonb we took these projects as already being part of the ppc64le distros from debian/ubuntu. So this doesn't represent a specific user or use case, but the simple fact that this is viewed as a base/common package for Linux at this point and we enable all of these packages for Power. Our goal is to hit full parity with Intel in terms of application availability. we can't name a specific django-waffle user, but usually ubuntu includes packages like this because they believe that they are the future or at least modern software and we piggy back on that decision. I know that's probably not an insightful "who are my users" answer that you are looking for but that's how we got here.

clintonb (approved):

PR #386, opened 2021-01-14
Fix flush with reverse m2m changed signal

k4rl85:
Fix issue https://github.com/django-waffle/django-waffle/issues/377

### DIFF
--- a/waffle/signals.py
+++ b/waffle/signals.py
@@ -8,4 +8,12 @@ from waffle import get_waffle_flag_model
 @receiver(m2m_changed, sender=get_waffle_flag_model().groups.through)
 def flag_membership_changed(sender, instance, action, **kwargs):
     if action in ('post_add', 'post_remove'):
-        instance.flush()
+        flag_model = get_waffle_flag_model()
+
+        # instance could be a flag or an instance of the related model
+        # https://docs.djangoproject.com/en/dev/ref/signals/#m2m-changed
+        if isinstance(instance, flag_model):
+            instance.flush()
+        else:
+            for flag in flag_model.objects.filter(pk__in=kwargs['pk_set']):
+                flag.flush()
--- a/waffle/tests/test_waffle.py
+++ b/waffle/tests/test_waffle.py
@@ -164,6 +164,29 @@ class WaffleTests(TestCase):
         response = process_request(request, views.flag_in_view)
         self.assertEqual(b'off', response.content)
 
+    def test_remove_from_user(self):
+        """Same operation of `test_user` but performed with reverse relation"""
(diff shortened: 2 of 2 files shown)


waffle/signals.py
```diff
@@ -8,4 +8,8 @@
 @receiver(m2m_changed, sender=get_waffle_flag_model().groups.through)
 def flag_membership_changed(sender, instance, action, **kwargs):
     if action in ('post_add', 'post_remove'):
-        instance.flush()
+        try:
+            instance.flush()
+        except AttributeError:
+            for flag in get_waffle_flag_model().objects.filter(pk__in=kwargs['pk_set']):
```
clintonb: Can you explain why this is necessary, please?  Is there anyway to do this without a database call?

k4rl85: Hi Clinton, first of all thx for the review on this patch.

The problem here is `m2m_changed` handling from Django.

If i execute the following code
```
from my_app.models import MyUserModel
from waffle.models import Flag

flag = Flag.objects.first()
user = MyUserModel.objects.first()

flag.users.add(user) # all fine, cache is flushed without errors

user.flag_set.add(flag) # crash
```

the last line will crash because the `instance` received by `flag_membership_changed` is a `MyUserModel` instance instead of a `Flag` instance.

The only solution that i can see  when `instance` is of type user/group is to retrieve all the flag involved and flush them like i do here

The reason why i query the db is that is not possible to guarantee that the flag involved are still accessible from `instance` when `flag_membership_changed` is executed.

Let me know if you want some of this explanation in the code.

I have added 2 tests to ensure that the `m2m_changed` signal is tested when the change are performed from the `User/Group` instance

k4rl85: I have changed the code to use `reverse` instead of `try/except` to be more explicit of what is happening here

clintonb: Thanks for the explanation. Would you mind adding a comment with a version of this explanation, too?

Also, instead of checking for a `kwarg`, what if you checked: `if isinstance(instance, get_waffle_flag_model())`? That seems a bit safer than relying on the `kwarg`.

clintonb (changes requested):

waffle/signals.py
k4rl85: I have applied the required changes.

clintonb (approved):

k4rl85:
Thx @clintonb for the merge.

What are the policy for this repository to have a new release? 

Unfortunately this bug impact my production code and i would avoid if possible to install django-waffle from github.

clintonb:
@k4rl85 new release is out.

PR #387, opened 2021-01-20
Moved to CircleCI

clintonb:


clintonb:
Going with GitHub Actions instead. See #389.

Issue #395, opened 2021-05-18
Test and release for Django 3.2


MaybeThisIsRu:
Hello!

Django 3.2 final was released last month. It would be great if this project is explicitly tested against and released with Django 3.2 compatibility.

PR #397, opened 2021-05-19
Add support for Django 3.2

### LINKED ISSUE #396: Default auto field not configured
[In Django 3.2](https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys), you can customise the type of auto-created primary key fields by configuring `DEFAULT_AUTO_FIELD` in your settings.

My project has recently been upgraded to Django 3.2, and I set `DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'`, because Django is making this the new default (previously it was `AutoField`). However, after doing that, the `makemigrations` command generates migrations to change this for dependencies that don't have a default configured, as well as first-party apps.

My project runs `makemigrations --check` in CI, and that now fails because it thinks waffle is missing a migration.

This could be addressed by adding `default_auto_field = 'django.db.models.AutoField'` to `WaffleConfig` - I'd be happy to make a PR to do that.

This also relates to issue #395

con-cat:
- Configure default auto field for waffle and test_app - Django 3.2 adds
  the ability to customise the type of auto-created primary key fields.
- Add Django 3.2 to tox.ini and setup.py

Addresses #395
Fixes #396

### DIFF
--- a/waffle/apps.py
+++ b/waffle/apps.py
@@ -4,6 +4,7 @@ from django.apps import AppConfig
 class WaffleConfig(AppConfig):
     name = 'waffle'
     verbose_name = 'django-waffle'
+    default_auto_field = 'django.db.models.AutoField'
 
     def ready(self):
         import waffle.signals  # noqa: F401
--- a/run.sh
+++ b/run.sh
@@ -17,17 +17,17 @@ shift
 
 case "$CMD" in
     "test" )
-        DJANGO_SETTINGS_MODULE=test_settings django-admin.py test waffle $@ ;;
+        DJANGO_SETTINGS_MODULE=test_settings django-admin test waffle $@ ;;
     "lint" )
         flake8 waffle $@ ;;
     "shell" )
-        django-admin.py shell $@ ;;
+        django-admin shell $@ ;;
     "makemigrations" )
-        django-admin.py makemigrations waffle $@ ;;
+        django-admin makemigrations waffle $@ ;;
     "makemessages" )
-        export DJANGO_SETTINGS_MODULE= && cd waffle && django-admin.py makemessages --all && cd - ;;
+        export DJANGO_SETTINGS_MODULE= && cd waffle && django-admin makemessages --all && cd - ;;
     "compilemessages" )
-        export DJANGO_SETTINGS_MODULE= && cd waffle && django-admin.py compilemessages && cd - ;;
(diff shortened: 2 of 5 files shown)


clintonb:
Since this feature pertains to Django 3.2, please implement #395, and update tox.ini and setup.py.

con-cat:
> Since this feature pertains to Django 3.2, please implement #395, and update tox.ini and setup.py.

@clintonb I've done that, and also updated `run.sh` to fix `RemovedInDjango40Warning: django-admin.py is deprecated in favor of django-admin.`

waffle/apps.py
```diff
@@ -4,6 +4,7 @@
 class WaffleConfig(AppConfig):
     name = 'waffle'
     verbose_name = 'django-waffle'
+    default_auto_field = 'django.db.models.AutoField'
```
MaybeThisIsRu, line 7: If I install `django-waffle` (once released with this change) and install it on Django 3.2, would it be in my interest to use `BigAutoField` even for this package, by manually configuring this?

Sorry if I'm getting something obvious very wrong here.

con-cat: You shouldn't need to change this - `AutoField` is still the default for projects that have been upgraded to 3.2, and the difference between the two is pretty minimal aside from `BigAutoField` giving a larger range of values when you get to very high ID numbers.

I left it as `AutoField` to keep things simple and avoid generating migrations, but I'd be happy to change to `BigAutoField` if people would prefer.

MaybeThisIsRu: That makes sense. I should have clarified I'm asking from the point of view of a fresh Django 3.2 project, rather than an upgrade.

con-cat: You shouldn't need to change it either way

con-cat:
I've removed the main branch of Django from the tox.ini envlist because of a few things that are incompatible with Django 4.0 causing tests to fail - I think it would make sense to address those separately.

Specifically, I tried replacing `ugettext_lazy` with `gettext_lazy` to address the error in the CI run above, but then I received the below error about a migration changing `NullBooleanField` to `BooleanField`, which is funny because NullBooleanField should still be ok in historical migrations.
```
ValueError: Cannot alter field waffle.Flag.everyone into waffle.Flag.everyone - they do not properly define db_type (are you using a badly-written custom field?)
```

clintonb (approved):

PR #405, opened 2021-07-21
Support setting testing flags with http headers

jsatt:
Using flag testing within a backend API doesn't work as well when you have to trigger the flag via a query params. I believe that using HTTP headers is a more natural way of indicating these flags for an API. This PR adds support for checking headers for `DWFT-Flag-Name` to enable/disable the flag the same as using the query params. This shouldn't affect the existing functionality, simply make Waffle more flexible for additional use cases.

### DIFF
--- a/waffle/tests/test_waffle.py
+++ b/waffle/tests/test_waffle.py
@@ -353,6 +353,22 @@ class WaffleTests(TestCase):
         assert not waffle.flag_is_active(request, 'foo')
         assert not hasattr(request, 'waffle_tests')
 
+    def test_testing_flag_header(self):
+        waffle.get_waffle_flag_model().objects.create(name='foo', testing=True)
+        request = RequestFactory().get('/foo', HTTP_DWFT_FOO='1')
+        request.user = AnonymousUser()
+        assert waffle.flag_is_active(request, 'foo')
+        assert 'foo' in request.waffle_tests
+        assert request.waffle_tests['foo']
+
+        # header should override cookie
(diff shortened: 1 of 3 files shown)


waffle/tests/test_waffle.py
```diff
@@ -353,6 +353,22 @@ def test_testing_disabled_flag(self):
...
 
+    def test_testing_flag_header(self):
+        waffle.get_waffle_flag_model().objects.create(name='foo', testing=True)
+        request = RequestFactory().get('/foo', HTTP_DWFT_FOO='1')
+        request.user = AnonymousUser()
+        assert waffle.flag_is_active(request, 'foo')
+        assert 'foo' in request.waffle_tests
+        assert request.waffle_tests['foo']
+
+        # header should override cookie
```
clintonb, line 364: Update the docs to list the order of preference for test flag values. If I have a cookie, header, and querystring set, which one wins?

clintonb (changes requested):
This looks great. We just need to update the docs to list the hierarchy of preferences.

waffle/tests/test_waffle.py
jsatt: @clintonb updated

clintonb (approved):

clintonb:
@jsatt fix the tests, and I'll re-approve.

jsatt:
@clintonb tests (and Django 2.2 support) should be fixed. Let's take another swing at it.

Issue #411, opened 2021-10-06
Migration `0004` blows up with Django 4 (NullBooleanField?)


hugorodgerbrown:
If you run the test suite against Django 4.0a1, it'll fail on setup with the following message:

```
ValueError: Cannot alter field waffle.Flag.everyone into waffle.Flag.everyone - 
they do not properly define db_type (are you using a badly-written custom field?)
```

The migration that fails is `0004_update_everyone_nullbooleanfield`, which, rather embarrassingly, was added by me when Django dropped the `NullBooleanField`. (See commit https://github.com/django-waffle/django-waffle/commit/4e365fef4e7c131ca4c344ef5751eeec77da9c86).

It looks like the problem is caused by going from a `NullBooleanField` (no longer supported) to a `BooleanField` - which occurs in migration 0004. This situation would not occur in a mature codebase, where this migration has already been applied, however it does appear in test runs where the database is being set up, and will occur for anyone who adds `django-waffle` if to a Django 4+ project.

I believe the solution is to retrospectively edit the existing migrations to remove the `NullBooleanField` references. I'll submit a PR just to check it passes CI.

mrxsal:
I am on Django==1.11.29, and when installing django-waffle, I now get the following error:
```
waffle.Flag.everyone: (fields.E110) BooleanFields do not accept null values.
        HINT: Use a NullBooleanField instead.
```

Could it be related to this change?

clintonb:
@mrxsal Django 1.x has not been supported for quite some time. We cannot help you with those errors. You’ll need to either upgrade Django, or use an older version of Django-Waffle.

mrxsal:
Thanks. django-waffle==1.0.0 works.

dancergraham:
The tests now pass with Django 4. I am closing this issue. Please open it again if you think this is wrong.

PR #412, opened 2021-10-06
WIP: Update migrations to remove NullBooleanField

hugorodgerbrown:
Creating this PR to run CI - WIP.

This commit retroactively edits the 0001 and 0003 migrations
to remove NullBooleanField references, as this field was
deprecated in Django 3, and is causing Django 4 tests to fail.

Fixes issue #411

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,7 @@
 [tox]
 envlist =
     py35-django22
-    py{36,37,38,39}-django{22,30,31,32}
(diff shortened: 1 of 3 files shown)


carltongibson (commented):
This shouldn't be necessary. [NullBooleanField is maintained as a stub for compatibility with historical migrations](https://github.com/django/django/blob/5111b636d9b63535fa8990142f3b5c756d08c1b6/django/db/models/fields/__init__.py#L1983-L2007).

PR #413, opened 2021-10-11
Refactor deprecated unittest aliases for Python 3.11 compatibility.

tirkarthi:
Deprecated unittest aliases were removed in Python 3.11 . This will cause error when tests are executed in Python 3.11 .

Ref : python/cpython#28268

assertEquals -> assertEqual
assertRaisesRegexp -> assertRaisesRegex


cleder:
Can you check with https://pypi.org/project/teyit/ if there are other suboptimal assertions used?

clintonb (approved):

PR #427, opened 2022-03-02
Make CREATE_MISSING_FLAGS logic universal (#400)

### LINKED ISSUE #400: CREATE_MISSING_FLAGS should be handled in .get() instead of is_active()
I am using the Flags in other scenarios aside from just views (celery tasks for example) and I was surprised that the `CREATE_MISSING_FLAGS` setting wasn't respected (I am manually calling `is_active_for_user` and other, custom, methods I created based on the [example](https://waffle.readthedocs.io/en/stable/types/flag.html#custom-flag-models)) and the methods were failing with error:
```
ValueError: "<CustomFlag: flag_name>" needs to have a value for field "id" before this many-to-many relationship can be used.
```

In my mind there should be a way to check if a flag is active using similar mechanism to `waffle.flag_is_active` which would work and respect the `CREATE_MISSING_FLAGS` setting, simplest solution seems to move the logic from `is_active` to `get()`. I would be happy to submit a PR if this would be considered a useful addition.

honzakral:
Before CREATE_MISSING_FLAGS was only respected when `is_active(request)` method was used, now any check against a flag (calling `get(name)`) will trigger the automatic creation of the `Flag` model.

Fixes #400

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -56,6 +56,13 @@ class BaseModel(models.Model):
         objects = cls.objects
         if get_setting('READ_FROM_WRITE_DB'):
             objects = objects.using(router.db_for_write(cls))
+
+        if get_setting('CREATE_MISSING_FLAGS'):
+            flag, _created = objects.get_or_create(
+                name=name, defaults={'everyone': get_setting('FLAG_DEFAULT')}
+            )
+            return flag
+
         return objects.get(name=name)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


clintonb (approved):

clintonb (changes requested):
@HonzaKral thanks for the speedy work. Can you add a test for this functionality?

honzakral:
@clintonb most of it was covered by existing tests so I just added a simple test that bypasses the current flow and would fail without the patch, hope that helps. If there is any other test you'd like to see I am happy to add it.

clintonb (approved):

PR #438, opened 2022-03-28
[WIP] Create missing model (Switch/Flag/Sample) in get_from_db

### LINKED ISSUE #400: CREATE_MISSING_FLAGS should be handled in .get() instead of is_active()

honzakral:
Fixes #400 

In my previous PR (#427) I made a mistake of ignoring `Switch` and `Sample` objects and just moving logic specific to `Flag` to `BaseModel`. This PR attempts a more generic solution that should work for all the mentioned models.

Marked as WIP due to missing tests

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -56,6 +56,13 @@ class BaseModel(models.Model):
         objects = cls.objects
         if get_setting('READ_FROM_WRITE_DB'):
             objects = objects.using(router.db_for_write(cls))
+
+        if get_setting('CREATE_MISSING_FLAGS'):
+            flag, _created = objects.get_or_create(
+                name=name, defaults={'everyone': get_setting('FLAG_DEFAULT')}
+            )
+            return flag
+
         return objects.get(name=name)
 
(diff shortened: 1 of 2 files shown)


clintonb:
Rebase and add tests when ready for review.

PR #444, opened 2022-06-01
Add config to support disabling admin pages by default.

onurhunce-planetly:
https://github.com/django-waffle/django-waffle/issues/443


docs/starting/configuring.rst
```diff
@@ -101,3 +101,6 @@ behavior.
     e.g. ``logging.WARNING``.
     Defaults to ``None``.
     
+
+``WAFFLE_ADMIN_PAGES_ARE_ENABLED``
```
clintonb: `WAFFLE_ENABLE_ADMIN_PAGES`

docs/starting/configuring.rst
```diff
@@ -101,3 +101,6 @@ behavior.
...
+    Admin pages can be disabled via this configuration. It is True by default.
```
clintonb: Enables the default admin pages for Waffle models. This is `True` by default, but can be disabled to override or customize the pages.

clintonb (changes requested):
I have suggestions for improving the naming and description. Otherwise, this is great!

docs/starting/configuring.rst
onurhunce-planetly: Done ✅

docs/starting/configuring.rst
onurhunce-planetly: Done ✅

clintonb (approved):

PR #446, opened 2022-06-19
Django 4.0 compatibility

triopter:
The test suite fails on Django 4.0 on trunk.  Fixes were pretty simple: two deprecated imports (one present only in the test app) had finally been removed, and support had been removed for the way the middleware was instantiated in the tests.

### DIFF
--- a/test_app/models.py
+++ b/test_app/models.py
@@ -1,7 +1,12 @@
 from django.contrib.auth.base_user import AbstractBaseUser
 from django.db import models
 from django.db.models import CASCADE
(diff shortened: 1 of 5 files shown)


clintonb (approved):

triopter:
I'm afraid I don't understand the release-test failure or how to fix it, but would be happy to address it if given some direction.

clintonb:
@triopter I need to update the credentials. The idea behind the test is to avoid surprises when publishing a release.

PR #452, opened 2022-07-22
Expose JSON endpoint for Waffle flag/switch/sample state

shaungallagher:
## Why

Waffle currently has a view to expose flag/switch/sample state as a Javascript object via the WaffleJS endpoint.

But there are a number of use cases where it would be preferable to expose that state in a serialized format like JSON.

(Related Issue: https://github.com/django-waffle/django-waffle/issues/449)

## What

Adds a JSON endpoint that returns the state of all flags/switches/samples.

In addition to the state, the endpoint also includes the last-modified time, which can be useful to identify stale flags that may need to be cleaned up.

## Quality

Adds unit tests for the new `waffle_status` view.

Adds documentation about how to expose and query the endpoint.

### DIFF
--- a/docs/usage/json.rst
+++ b/docs/usage/json.rst
@@ -1,0 +1,68 @@
+.. _usage-json:
+
+=====================
+Waffle Status as JSON
+=====================
+
+Although :doc:`WaffleJS<javascript>` returns the status of all
+:ref:`flags <types-flag>`, :ref:`switches <types-switch>`, and
+:ref:`samples <types-sample>`, it does so by exposing a Javascript
+object, rather than returning the data in a directly consumable format.
+
+In cases where a directly consumable format is preferrable,
+Waffle also exposes this data as JSON via the ``waffle_status`` view.
+
+
+Using the view
+--------------
+
+Using the ``waffle_status`` view requires adding Waffle to your URL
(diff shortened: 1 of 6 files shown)


docs/usage/json.rst
```diff
@@ -0,0 +1,68 @@
...
+        (r'^', include('waffle.urls')),
+    )
+
+This adds a route called ``waffle_status``, which will return the current
+status of each flag, switch, and sample as JSON, with the following structure:
+
+.. code-block:: json
+
+    {
+        "flags": [
```
clintonb: When I use flags, I generally only care about  1-2 of them. What if you returned an object, where the key is the name and the value is the object containing the status and timestamp, instead of an array?

```json
"flags": {
  "flag_active": {
    "is_active": true,
    "last_modified": "2020-01-01T12:00:00.000"
  }
}
```

This would make accessing the data simpler for end users.

clintonb (commented):
This all looks fine. Let me know your thoughts on the data format.

docs/usage/json.rst
shaungallagher: I like that a lot.  Will update.

clintonb (approved):

Issue #456, opened 2022-08-22
Auto-create Custom Switch Models


cole-jacobs:
Auto-creating custom switches fails with: `AttributeError: Manager isn't available; 'waffle.Switch' has been swapped for '<custom model>'`.

In models.py, the Flag version of `is_active` calls `get_waffle_flag_model` before its call to `get_or_create`. I think a similar pattern should be used for switches and samples. Happy to submit a patch for this tonight.

clintonb:
Yes, please patch.

cole-jacobs:
👌 I'll get something up soon. Just noticed the same idea can be added to `override_switch` and `override_sample`, so I'll include that as well.

PR #457, opened 2022-08-23
Pluggable Switch and Sample models #456

cole-jacobs:
Remove direct references to `Switch` and `Sample` in favor of
`get_waffle_switch_model` and `get_waffle_sample_model`.

### DIFF
--- a/waffle/management/commands/waffle_sample.py
+++ b/waffle/management/commands/waffle_sample.py
@@ -1,7 +1,6 @@
 from django.core.management.base import BaseCommand, CommandError
 
-from waffle.models import Sample
-
+from waffle import get_waffle_sample_model
 
 class Command(BaseCommand):
     def add_arguments(self, parser):
@@ -30,7 +29,7 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         if options['list_samples']:
             self.stdout.write('Samples:')
-            for sample in Sample.objects.iterator():
+            for sample in get_waffle_sample_model().objects.iterator():
                 self.stdout.write('%s: %.1f%%' % (sample.name, sample.percent))
             self.stdout.write('')
             return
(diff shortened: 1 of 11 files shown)


waffle/management/commands/waffle_sample.py
```diff
@@ -1,7 +1,6 @@
+from waffle import get_waffle_sample_model
 from django.core.management.base import BaseCommand, CommandError
 
-from waffle.models import Sample
```
clintonb, line 3: Keep the local imports separate from third-party.

waffle/management/commands/waffle_switch.py
```diff
@@ -1,8 +1,7 @@
 from argparse import ArgumentTypeError
+from waffle import get_waffle_switch_model
```
clintonb: Keep the local imports separate from third-party.

waffle/tests/test_views.py
```diff
@@ -23,14 +26,17 @@ def test_waffle_status(self):
         self.assertIn('max-age=0', cache_control)
 
     def test_waffle_status_response(self):
-        get_waffle_flag_model().objects.create(name='test_flag_active', everyone=True)
```
clintonb: It doesn't seem anything actually needed to be changed here. Revert changes to this section.

clintonb (changes requested):

cole-jacobs:
What's going on with the automated tests? I ran the test suite locally and everything passed 🤔

clintonb (approved):

cole-jacobs:
Thanks @clintonb for approving my changes. Anything left to do before merge and release?

waffle/tests/test_views.py
clintonb: Again, just revert everything in this section since it's not necessary. Line 33 is the source of your test failure.

clintonb (changes requested):

waffle/tests/test_views.py
cole-jacobs: Ahhhh, I see it now. Sorry about that.

clintonb (approved):

PR #458, opened 2022-08-24
Remove EOL Python & Django versions

ulgens:


tox.ini
```diff
@@ -1,23 +1,17 @@
 [tox]
 envlist =
-    py35-django22
-    py{36,37,38,39}-django{22,30,31,32}
+    py{37,38,39}-django{32}
     py{310}-django{32,40}
```
clintonb, line 4: Add Django 4.1

clintonb:
Duplicated by https://github.com/django-waffle/django-waffle/pull/460

PR #466, opened 2022-09-07
Add python_requires and install_requires to help pip

gopackgo90:
Without `python_requires`, pip will let you install django-waffle 3.0 on python 3.6 and lower, which are now unsupported. And django-waffle should depend on django, because based on the name of the project, it does, so adding `install_requires` to reflect that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,8 @@ setup(
     include_package_data=True,
     package_data={'': ['README.rst']},
     zip_safe=False,
+    python_requires='>=3.7',
+    install_requires=['django>=3.2'],
     classifiers=[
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -13,6 +13,8 @@
     include_package_data=True,
     package_data={'': ['README.rst']},
     zip_safe=False,
+    python_requires='>=3.7',
+    install_requires=['django>=3.2']
```
clintonb: Add a comma

clintonb (changes requested):

clintonb (approved):

PR #480, opened 2022-12-02
feat: Allow override_flag testutil to set different checks

### LINKED ISSUE #439: Test util needed to enable flag for specific user instead of all users
In automated tests, to correctly test features sometimes a flag needs to be enabled for one user but not enabled for another user. The current test helper `override_flag` is can only mock on for all users or off for all users.

adamantike:
Up until now, the `override_flag` test util was only useful to completely activate or deactivate a flag. It didn't allow to provide more customization for all the internal checks it makes, like evaluating if the request user is authenticated or staff.

This change provides that functionality, without breaking the current API, to avoid breaking changes. Now, users can provide more granular arguments to `override_flag`, for a more flexible flag testing.

Closes #439.

### DIFF
--- a/docs/testing/automated.rst
+++ b/docs/testing/automated.rst
@@ -41,6 +41,17 @@ Or::
         # samples behave normally.
         assert waffle.sample_is_active('sample_name')
 
+``override_flag`` also allows providing values for its different checks.
+For example::
+
+    @override_flag('flag_name', percent=25)
+    def test_with_flag():
+        ...
+
+    @override_flag('flag_name', staff=True, superusers=True)
+    def test_with_flag():
+        ...
+
 All three will restore the relevant flag, sample, or switch to its
 previous state: they will restore the old values and will delete objects
 that did not exist.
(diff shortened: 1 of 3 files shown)


clintonb (commented):
What's the use case for this? It seems the user of this functionality will ultimately either test waffle or the test itself.

For example, what does it mean to enable a flag for 25% in an automated test? Typically you want to test behavior when the flag is active vs. inactive. Setting a 25% rollout will result in flaky tests. This will be a footgun for potential users.

That user override use case is solved by instantiating the flag in the test: `waffle.get_waffle_flag_model().objects.create(name='foo', everyone=None, superusers=True)`. However, flags are created either by migration or manually. If you create them in your test, you're really only testing your tests. If you run migrations in your test to setup the flag, your tests don't need overrides.

Flagged code doesn't care about the conditions that resulted in the flag's active state, but the state itself.

clintonb:
Closing due to inactivity.

PR #484, opened 2023-04-13
Django 4.2.0 and Python 3.11 compatibility

### LINKED ISSUE #483: Is this still active?
This project was very active for over 10 years then it just stopped.  Does anyone know why?  Thinking about using this, but the stop is giving me pause.

bennylope:
This PR doesn't change any of the underlying library code, it just runs the tests against Django 4.2 and Python 3.11 (for supported Django versions, 4.1 and 4.2).

The [commit adding `run.sh` to `allowlist_externals`](1f18fcd82b215975fec8fb05c70fffe5af8b7b40) was required to get the tests running under `tox` and given that it's something I've had to implement before figured it was probably important enough to include. Though it is surprising that it's not been an issue for anyone else... 

This should presumably be enough to close #483 and #411

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ 3.7, 3.8, 3.9, '3.10' ]
+        python-version: [ 3.7, 3.8, 3.9, '3.10', '3.11' ]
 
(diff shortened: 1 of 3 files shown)


clintonb (approved):

irtazaakram:
Python 3.7 has reached its EOL https://devguide.python.org/versions/#unsupported-versions.

Please can we have new release after this PR is merged.

PR #491, opened 2023-11-23
add support for everyone option to is_active_for_user

### LINKED ISSUE #401: is_active_for_user() doesn't respect flag.everyone
This was surprising to me – `is_active_for_user()` is exposed as a public API function, so I thought it would handle `flag.everyone`, but it doesn't respect that setting. That makes sense to me for `_is_active_for_user()` (which happens after the check on `.everyone` in `is_active()`, but makes for unexpected behavior when trying to use flags outside of a request context (where `is_active_for_user()` feels like the right API to use...

I would expect there to be a simple way to check if a flag is active without a request (e.g. in a celery task) but instead I have to ensure that I check the `.everyone` field separately. Is this intended? Could the API be changed so that `is_active_for_user()` resepects `.everyone`?

jdoconnor:
fixes #401 

This adds support of the `everyone` boolean to `is_active_for_user`

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -230,6 +230,9 @@ class AbstractBaseFlag(BaseModel):
         return flush_keys
 
     def is_active_for_user(self, user: AbstractBaseUser) -> bool | None:
+        if self.everyone:
+            return True
+
         if self.authenticated and user.is_authenticated:
             return True
 
--- a/waffle/tests/test_models.py
+++ b/waffle/tests/test_models.py
@@ -5,6 +5,7 @@ from waffle import (
     get_waffle_sample_model,
     get_waffle_switch_model,
 )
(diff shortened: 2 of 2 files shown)


clintonb (approved):

jwhitlock:
Should this also return `False` if `everyone` is `False`, and only continue on a per-user check if `everyone` is `None`? This is how `flag.is_active` works:

https://github.com/django-waffle/django-waffle/blob/63444d6f83a3c75a9630a02e32cfe65ac1305b86/waffle/models.py#L282-L285

chess-octane:
I agree with @jwhitlock . I wonder what are your thoughts @clintonb ?

chess-octane:
I submitted a PR to tackle that: https://github.com/django-waffle/django-waffle/pull/498

Issue #492, opened 2023-11-29
Command to remove unused switches/flags


cristianowa:
In our usage of django-waffle when switches are removed they often are left in at least one environment database without any meaningful use. 

The cost it self is irrelevant for a DB, but checking for active switches (usually debugging a production issue) get more complex overtime.

I have made this script  to clean the switches and executed it on Django shell.

```
from waffle.models import *
switches = Switch.objects.all()

from subprocess import getoutput
from subprocess import getstatusoutput as cmd

print("name", "is_active", "created", "modified")
for switch in switches:
    switch_not_found, _ = cmd("grep -ri {switch.name} *")
    if switch_not_found:
       print(switch.name, switch.is_active, switch.created, switch.modified)
       switch.delete()

```

I saw there is [already a command](https://waffle.readthedocs.io/en/stable/usage/cli.html?highlight=remove%20unused%20switches#deleting-data) for deleting data.

My proposal here is to create a command like `waffle_remove_unused` with options for switches, flags and samples, and also having a dry-run (just printing what would be deleted).  

If that is OK I would be glad to work on a PR for it.

The first thing being changing that grep to a Python implementation. Second is to make this generic to Samples and Flags as well.

clintonb:
A pull request is welcome. I, personally, would be hesitant to use such a command due to lack of trust. It's far simpler for me to grep/search the code myself than hope someone else's command does the right thing. But, that's me.

cristianowa:
Got it @clintonb, thanks for the feedback. I think the command should not be used in all cases, but in some it makes sense. In our case we have a system with hundreds of feature flags, and a lot of them were forgotten/lost in the process of removal. 

I will work on the code and will open the PR as soon as I can.

cristianowa:
@clintonb When you have the time, please share you thoughts on #493 . :slightly_smiling_face:

levimoore1992:
Any update on this?

PR #493, opened 2023-12-07
Create command to remove unused data by checking existing code

cristianowa:
This PR is related to the issue #492 .

The idea behind it is that in some cases the switches/flags/samples can be automatically removed by checking if they are present in the source code.

Parsing the code for that is implemented directly in Python using the standard lib to maximize compatibility. 

The default behavior is to ask for a confirmation in each flag, but there is an override.

I've also checked the messages `./run.sh makemessages` and there were no new ones, so I didn't updated those files.

### DIFF
--- a/waffle/management/commands/waffle_delete_unused.py
+++ b/waffle/management/commands/waffle_delete_unused.py
@@ -1,0 +1,84 @@
+import os
+import pathlib
+
+from django.core.management.base import BaseCommand
+from waffle import (
+    get_waffle_flag_model,
+    get_waffle_switch_model,
+    get_waffle_sample_model,
+)
+
+
+class Command(BaseCommand):
+    help = "Delete flags, samples, and switches not present in the code from the Database"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Do not delete anything, just show what would be deleted",
+        )
+        parser.add_argument(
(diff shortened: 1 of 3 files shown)


waffle/management/commands/waffle_delete_unused.py
```diff
@@ -0,0 +1,84 @@
...
+def expression_in_file(expression, filename):
+    with open(filename) as file:
+        content = file.read()
+        return expression in content
+
+
+def expression_exists(expression):
+    for root, dirs, files in os.walk(os.getcwd()):
+        for file in files:
+            if not (file.endswith(".py") or file.endswith(".html")): #TODO: make this a list of extensions
```
clintonb, line 79: Are you going to do this? Perhaps allow them to be passed in as an optional argument?

waffle/tests/test_management.py
```diff
@@ -329,3 +329,44 @@ def test_delete_some_but_not_all_records(self):
 
         call_command('waffle_delete', flag_names=[flag_1])
         self.assertTrue(get_waffle_flag_model().objects.filter(name=flag_2).exists())
+
+
+class WaffleDeleteUnused(TestCase):
```
clintonb, line 334: These tests aren't really testing the desired behavior—deleting flags. Consider updating the command to take a file path of code to search, and passing that in the test calls to confirm items are deleted.

PR #498, opened 2024-02-20
Make is_active_for_user consider when everyone is False

### LINKED ISSUE #401: is_active_for_user() doesn't respect flag.everyone

chess-octane:
This adds a check to the `is_active_for_user` method to consider when the attribute `everyone` is False.

Fixes #401

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -233,6 +233,9 @@ class AbstractBaseFlag(BaseModel):
         if self.everyone:
             return True
 
+        if self.everyone is False:
+            return False
+
         if self.authenticated and user.is_authenticated:
             return True
 
@@ -396,7 +399,8 @@ class AbstractUserFlag(AbstractBaseFlag):
 
     def is_active_for_user(self, user: AbstractBaseUser) -> bool | None:
         is_active = super().is_active_for_user(user)
-        if is_active:
+
+        if is_active is True or is_active is False:
             return is_active
 
         user_ids = self._get_user_ids()
--- a/waffle/tests/test_models.py
+++ b/waffle/tests/test_models.py
@@ -1,3 +1,4 @@
+from django.contrib.auth import get_user_model
 from django.test import TestCase
 
 from waffle import (
@@ -5,8 +6,6 @@ from waffle import (
     get_waffle_sample_model,
     get_waffle_switch_model,
 )
-from django.contrib.auth.models import User
-
 
 class ModelsTests(TestCase):
     def test_natural_keys(self):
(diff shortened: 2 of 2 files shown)


waffle/models.py
```diff
@@ -233,6 +233,9 @@ def is_active_for_user(self, user: AbstractBaseUser) -> bool | None:
         if self.everyone:
             return True
 
+        if self.everyone is False:
```
renvieir, line 236: This and the previous if, could be changed to?
```suggestion
        if self.everyone != None:
            return self.everyone
```

waffle/models.py
```diff
@@ -396,7 +399,8 @@ def _get_group_ids(self) -> set[Any]:
 
     def is_active_for_user(self, user: AbstractBaseUser) -> bool | None:
         is_active = super().is_active_for_user(user)
-        if is_active:
+
+        if is_active is True or is_active is False:
```
renvieir, line 403: same here

waffle/models.py
chess-octane: this does not follow PEP standards

waffle/models.py
chess-octane: this is more readable

minhtrangvy: would `if is_active is not None` work the same? it is also more succinct and clear

minhtrangvy (approved):

waffle/models.py
```diff
@@ -233,6 +233,9 @@ def is_active_for_user(self, user: AbstractBaseUser) -> bool | None:
         if self.everyone:
```
minhtrangvy, line 233: maybe want to make this `if self.everyone is True`?

minhtrangvy (changes requested):
actually we have a usecase where we rely on `is_active_for_user` NOT respecting the "everyone" attribute in order to create feature flags that are on just for some users and not all

maybe this should be a required param on whether to respect `everyone` or not, otherwise i could see a lot of issues when people upgrade waffle because of this behavior change

waffle/models.py
chess-octane: the problem is that the result of `not None` is `True` which will not capture the `False` use case

joshkersey:
I disagree with @minhtrangvy. This is a bug in the code. The documentation makes it clear what the expectation for the `Everyone` attribute is:
`Everyone: | Globally set the Flag, overriding all other criteria. Leave as Unknown to use other criteria.`

PR #501, opened 2024-03-18
Implemented Jazzband guidelines 

clintonb:
See https://github.com/jazzband/django-waffle/issues/500


dragon-dxw (approved):
Would be nice to have a proper pre-commit config, but that can wait.

PR #504, opened 2024-05-03
Add flake8 to precommit

dragon-dxw:
We only check the waffle directory to keep it equivalent to the existing linter.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,1 +1,6 @@
-repos: []
+files: ^waffle/
+repos:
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -1 +1,6 @@
-repos: []
+files: ^waffle/
+repos:
+-   repo: https://github.com/pycqa/flake8
+    rev: '7.0.0'
+    hooks:
+    -   id: flake8
```
clintonb: Add a newline.

clintonb (approved):

Issue #80, opened 2013-05-07
Cookie is only returned for a testing flag set via GET param if that flag is queried during the request


matthewryanscott:
Given:
- `myflag` is set up with `testing=True`
- Page A never calls `waffle.flag_is_active(request, 'myflag')`, and none of the templates it loads have `{% flag myflag %}`
- Page B makes that function call, and/or loads a template with that tag.
- Browser has either `dwft_myflag` cookie as `False`, or does not have that cookie.

Then (current behavior):
- Request page A with GET param `?dwft_myflag=1`
- `dwft_myflag` cookie is never set to `True`
- Request page B without GET param, and `myflag` is treated as `False`

Desired behavior:
- Request page A with GET param `?dwft_myflag=1`
- `dwft_myflag` cookie is set to `True`
- Request page B without GET param, and `myflag` is treated as `True`

stefanfoulis:
this has bitten me as well

benje:
:+1: a fix for this would be very useful for me too (I wrongly assumed this would be default behaviour)

jsocol:
Honestly, I don't have a good idea of how to fix this. I have a couple of lousy ideas. The problem is that the prefix (`dwft_`) is user-configurable—in fact **it doesn't even have to be a prefix**.

Therein lies the problem. So, a couple of things that could work:
- break backward compat, make it necessarily a prefix
- do something very gross with regex, like `re.compile('^' + test_cookie_template.replace('%s', '(.*)') + '$')`

Of those, I'd rather break backward compat.

PR #505, opened 2024-05-03
Always set testing cookies, even on pages which do not use them

dragon-dxw:
Addresses https://github.com/jazzband/django-waffle/issues/80

Uses code from https://github.com/OpenHumans/open-humans/pull/1109

Modifies the tests slightly to attempt to set the cookie on a page which doesn't make use of waffle.


dragon-dxw:
Taking another look, I think there's a slight issue with my changes to the tests -- we no longer test whether the results of the flag toggle are instantly available at the instant of having been run. If people make good noises I'll fix that (by broadly replicating the original test)

PR #511, opened 2024-11-06
Add Django 5 and python 3.12 and 3.13 test support

dancergraham:
based on https://github.com/jazzband/django-waffle/pull/484/commits

PR #512, opened 2024-11-08
Create dependabot config for github-actions

ulgens:
Coming here from https://github.com/jazzband/django-waffle/pull/511 . I didn't want to apply actions updates by hand when they can be automatized, so there is the dependabot config. After the merge, triggering the updates will create an update PR automatically.


dancergraham:
Looks like a much better solution than manually updating them every time ;) I will be interested to see if this works - we have a similar config at my work and it fails silently every week!

dancergraham (approved):

clintonb (approved):

Issue #489, opened 2023-07-27
ruff: Two flake8-django issues to be fixed


cclauss:
# Good First Issue
% [`pipx install ruff `](https://pypi.org/project/pipx)
% [`ruff --select=DJ --show-source .`](https://beta.ruff.rs/docs/rules/#flake8-django-dj)
```
test_app/models.py:12:7: DJ008 Model does not define `__str__` method
   |
12 | class Company(models.Model):
   |       ^^^^^^^ DJ008
13 |     name = models.CharField(
14 |         max_length=100,
   |

waffle/models.py:107:5: DJ012 Order of model's inner classes, methods, and fields does not follow
    the Django Style Guide: `save` method should come before custom method
    |
105 |           cache.delete_many(keys)
106 |
107 |       def save(self, *args: Any, **kwargs: Any) -> None:
    |  _____^
108 | |         self.modified = timezone.now()
109 | |         ret = super().save(*args, **kwargs)
110 | |         if hasattr(transaction, 'on_commit'):
111 | |             transaction.on_commit(self.flush)
112 | |         else:
113 | |             self.flush()
114 | |         return ret
    | |__________________^ DJ012
115 |
116 |       def delete(self, *args: Any, **kwargs: Any) -> tuple[int, dict[str, int]]:
    |

Found 2 errors.
```
---
% [`ruff rule DJ008`](https://beta.ruff.rs/docs/rules/django-model-without-dunder-str/#django-model-without-dunder-str-dj008)
# django-model-without-dunder-str (DJ008)

Derived from the **flake8-django** linter.

## What it does
Checks that `__str__` method is defined in Django models.

## Why is this bad?
Django models should define `__str__` method to return a string representation
of the model instance, as Django calls this method to display the object in
the Django Admin and elsewhere.

Models without `__str__` method will display a non-meaningful representation
of the object in the Django Admin.

## Example
```python
from django.db import models


class MyModel(models.Model):
    field = models.CharField(max_length=255)
```

Use instead:
```python
from django.db import models


class MyModel(models.Model):
    field = models.CharField(max_length=255)

    def __str__(self):
        return f"{self.field}"
```
---
% [`ruff rule DJ012`](https://beta.ruff.rs/docs/rules/django-unordered-body-content-in-model/#django-unordered-body-content-in-model-dj012)

Derived from the **flake8-django** linter.

## What it does
Checks for the order of Model's inner classes, methods, and fields as per
the [Django Style Guide].

## Why is this bad?
[...]

PR #488, opened 2023-07-27
Use ruff to lint Python code

cclauss:
[Codespell](https://github.com/codespell-project/codespell#readme) can find and fix common misspellings in text files. It’s designed primarily for checking misspelled words in source code, but it can be used with other files as well.

[Ruff](https://beta.ruff.rs/) supports [over 600 lint rules](https://beta.ruff.rs/docs/rules) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

`ruff --output-format=github .` provides intuitive GitHub Annotations to contributors...

image

Related to
* #489

PR #518, opened 2024-11-09
Use ruff to lint Python code

cclauss:
[Ruff](https://docs.astral.sh/ruff) supports [over 800 lint rules](https://docs.astral.sh/ruff/rules/) and can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [yesqa](https://github.com/asottile/yesqa), [eradicate](https://pypi.org/project/eradicate/), [pyupgrade](https://pypi.org/project/pyupgrade/), and [autoflake](https://pypi.org/project/autoflake/), all while executing (in Rust) tens or hundreds of times faster than any individual tool.

`ruff check --output-format=github` provides intuitive GitHub Annotations to contributors...

image

Related to
* #488
* #489

@ulgens


pyproject.toml
```diff
@@ -62,3 +62,90 @@ strict_equality = true
...
+# Files not checked:
+#  - migrations: most of these are autogenerated and don't need a check
+#  - docs: contains autogenerated code that doesn't need a check
+exclude = [
+    "*/migrations/*",
+    "docs",
+]
+ignore = ["F401"]
+
+[tool.ruff.lint.mccabe]
```
dancergraham, line 138: question: why complexity 23 ?

pyproject.toml
```diff
@@ -62,3 +62,90 @@ strict_equality = true
 [[tool.mypy.overrides]]
 module = ["django.*"]
 ignore_missing_imports = true
+
+[tool.ruff]
+line-length = 120
+target-version = "py38"
+
+[tool.ruff.lint]
```
dancergraham, line 70: question: why this set of lint rules?

waffle/templatetags/waffle_tags.py
```diff
@@ -41,16 +41,15 @@ def render(self, context):
     def handle_token(cls, parser, token, kind, condition):
         bits = token.split_contents()
         if len(bits) < 2:
-            raise template.TemplateSyntaxError("%r tag requires an argument" %
-                                               bits[0])
+            raise template.TemplateSyntaxError(f"{bits[0]!r} tag requires an argument")
```
dancergraham, line 44: suggestion: use `repr(bits[0])` for readability in the spirit of https://peps.python.org/pep-0498/#s-r-and-a-are-redundant

dancergraham (commented):
ruff is a great tool. This allows dev workflows with constant linting. It makes no significant difference to ci times as flake 8 runs in parallel with the tests and finishes first.

Django uses flake8 but I don't see a good reason why this library needs to wait for django to move

I've looked through the changes and can't see anything which should break.

dancergraham:
I think the pre-commit.yaml file needs to be updated as well

pyproject.toml
cclauss: Set it to 22 and see what happens.

pyproject.toml
cclauss: These are the ones that currently pass.  Future PRs can uncomment rules that might be of interest.  We do not want this PR to contain so many changes that it slows the review process.

waffle/templatetags/waffle_tags.py
cclauss: This PR is about automatic changes, not manual changes.  Turning on different rules may make other automatic changes.

pyproject.toml
cclauss: Set it to 22 and see what happens...

% `pipx run ruff check --config mccabe.max-complexity=22`

pyproject.toml
```diff
@@ -62,3 +62,90 @@ strict_equality = true
...
+    "docs",
+]
+ignore = ["F401"]
+
+[tool.ruff.lint.mccabe]
+max-complexity = 23
+
+[tool.ruff.lint.per-file-ignores]
+"docs/conf.py" = ["INP001"]
+"test_app/models.py" = ["DJ008"]  # FIXME
```
dancergraham, line 143: praise: nice to highlight this in the comment

pyproject.toml
```diff
@@ -62,3 +62,90 @@ strict_equality = true
...
+"waffle/models.py" = ["DJ012", "PYI019"]  # FIXME
+
+[tool.ruff.lint.pylint]
+allow-magic-value-types = ["float", "int", "str"]
+max-args = 6  # default is 5
```
dancergraham, line 148: praise: nice to have the defaults mentioned here as a possible future target for refactoring

pyproject.toml
```diff
@@ -62,3 +62,90 @@ strict_equality = true
...
+  "DTZ",    # flake8-datetimez
+  "E",      # pycodestyle errors
+  "F",      # Pyflakes
+  "FIX",    # flake8-fixme
+  "FLY",    # flynt
+  "G",      # flake8-logging-format
+  "ICN",    # flake8-import-conventions
+  "INP",    # flake8-no-pep420
+  "INT",    # flake8-gettext
+  "NPY",    # NumPy-specific rules
```
dancergraham, line 86: nit: this library doesn't use numpy or pandas and seems unlikely to start doing so so these could be omitted for a minor gain in readability

dancergraham (commented):
I'm new around here so reluctant to personally approve such a big PR but it looks OK to me

clintonb (approved):

pyproject.toml
cclauss: > this library doesn't use numpy or pandas and seems unlikely to start doing so

Famous last words.  You never know when code will be forked or when contributors will have new ideas.  Given that ruff can lint the entire CPython codebase in less than a third of a second with no cache, I take the stance that if a family of rules pass then run them.

PR #519, opened 2024-11-09
Add Django 5.0 support

ulgens:


cclauss:
Why not dj5.1 too?

dancergraham (approved):
Looks good - I would add Django 5.1 at the same time as it is the currently supported version and the tests pass fine but 5.0 is better than 4.1. I would also remove 3.2, 4.0 and 4.1 as they are no longer supported 
https://www.djangoproject.com/download/#supported-versions

dancergraham:
suggestion: add a comment linking to the supported versions table to help with future updates
https://www.djangoproject.com/download/#supported-versions

cclauss:
As discussed in #516, we should keep adding support and removing support in two different pull requests.  Adding support is easy but removing support should be done on a serious version change so we do not break apps that use this software.

I prefer this URL for explaining the matrix of support:
* https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django

dancergraham:
> I prefer this URL for explaining the matrix of support:
> 
> * https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django

ahh yes much better - that explains the matrix between python and Django versions !

ulgens:
> Why not dj5.1 too?

I was trying to handle each version by their own PR; otherwise, the project's history would become too complicated and not useful for further contributions.

> Looks good - I would add Django 5.1 at the same time as it is the currently supported version and the tests pass fine but 5.0 is better than 4.1. I would also remove 3.2, 4.0 and 4.1 as they are no longer supported

Same reasoning here 🌷

The test matrix will get simplified when we start removing EOL versions.

dancergraham:
> I was trying to handle each version by their own PR; otherwise, the project's history would become too complicated and not useful for further contributions.

Maybe separate commits with thematically grouped PRs would be my preferred balance between time spent writing and reviewing minor updates but I'm fine with this approach too.

cclauss (approved):

PR #520, opened 2024-11-09
Add Django 5.1 support

ulgens:


cclauss (approved):

cclauss (changes requested):
Please add Python 3.13 to the testing because Django v5.1 adds support for it.
* https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django

ulgens:
@cclauss Purpose of these PRs are adding these versions individually. 3.13 will be a separate one.

cclauss:
Testing Django v5.1 should cover all versions of Python that it supports.

dancergraham (approved):

PR #523, opened 2024-11-09
Add support for Python 3.13

ulgens:
With this PR, all modern versions of Python and Django is tested & supported.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ 3.8, 3.9, '3.10', 3.11, 3.12 ]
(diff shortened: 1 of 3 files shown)


ulgens:
@clintonb What do you think about removing EOL versions? Do you want to make a release before removing them, with the new supported versions or just dive in to cleaning?

dancergraham:
My 2c I'd say there's no time like the present:
Frugality and environmental concern pushes us to remove unneeded ci jobs
Caring about users makes us think twice...
I don't know of a formal policy on this for this project or for Jazzband
The closest I can see for Jazzband is "don't be dogmatic"
Semantic versioning doesn't link version numbers to supported versions but rather to breaking API changes. 
People using EOL versions should be pinning minor versions of their dependencies

dancergraham (approved):

cclauss (approved):

Issue #524, opened 2024-11-13
documentation building fails


dancergraham:
Hello, the documentation build seems to be linked but failing on read the docs
https://app.readthedocs.org/projects/waffle/

The logs suggest that it is due to a missing `.readthedocs.yaml` configuration file 

https://app.readthedocs.org/projects/waffle/builds/26232926/

The previous successful builds seemed to use a `docs/conf.py` file for configuration

https://app.readthedocs.org/api/v2/build/21409878.txt

Theory [Edited]: conf.py is no longer sufficient, we need to add a .yaml file and probably migrate some settings from conf.py.

https://docs.readthedocs.io/en/stable/config-file/index.html

dancergraham:
fixed by #525

PR #525, opened 2024-11-13
Create .readthedocs.yaml for documentation configuration

dancergraham:
to fix broken documentation building pipeline

#524


dancergraham:
This **Should** work - I have basically just added the standard config file

I see no obvious way to test it without merging

That seems a low-risk way to test it as build is currently breaking so it can't get any worse!

clintonb (approved):

PR #528, opened 2024-11-16
Drop support for EOL Django versions (3.2, 4.0, 4.1)

ulgens:


clintonb (commented):
Move the commit hook, and let’s discuss separately. I’m unclear on the implications of this, and don’t want to unnecessarily burden folks just trying to fix bugs.

.pre-commit-config.yaml
```diff
@@ -9,3 +9,9 @@ repos:
     rev: v0.7.3
     hooks:
       - id: ruff
+
+-   repo: https://github.com/adamchainz/django-upgrade
```
dancergraham: Django upgrade documentation points out that a check for missing migrations is needed with this - do we have one?

dancergraham (commented):
I'm in favour - note that based on my reading of jazzband guidelines this does not need a major version bump.

.pre-commit-config.yaml
dancergraham: We use a test like this in our test suite at work 

```python
import unittest

from django.apps.registry import apps
from django.db import connection
from django.db.migrations.autodetector import MigrationAutodetector
from django.db.migrations.executor import MigrationExecutor
from django.db.migrations.state import ProjectState

class MigrationsTest(unittest.TestCase):
    def test_missing_migrations(self):
        executor = MigrationExecutor(connection)
        autodetector = MigrationAutodetector(
            executor.loader.project_state(),
            ProjectState.from_apps(apps),
        )
        changes = autodetector.changes(graph=executor.loader.graph)
        self.assertEqual({}, changes)

clintonb: Please add a test for missing migrations in a follow up PR. 

My concern with Django upgrades is not versioning, which I _do_ believe should be major versions if we are dropping support for something, but with the activation energy needed to make a change. If someone just wants to fix a small issue, or add a feature, I don’t want to burden them with Django compatibility issues. Again, I’m not familiar with this pre-commit hook, so I’m not comfortable merging it at this time. I am happy to discuss on a separate PR.

ulgens: @dancergraham 

> Django upgrade documentation points out that a check for missing migrations is needed with this - do we have one?

Can you please share a link for this? The abstraction layer this tool works is not dependent to migrations.

@clintonb I'll be sending the hook in a different PR, we can continue to discuss there.

dancergraham: Yes it's in the readme on GitHub

ulgens: @dancergraham Thank you. I wouldn't say missing migrations directly relates to django upgrade, it's just a note so people can take full benefit of the tool (which can be told about almost any of the best practices with Django). It's useful if the project is getting model changes often, but if you consider this repo is a library and the latest migration was 4 years ago, I don't think it will beneficial here.

ulgens: @clintonb Pre-commit hook is dropped now.

clintonb (approved):

PR #530, opened 2024-11-16
Update documentation

dancergraham:
Remove references to projects which are no longer maintained or to long abandoned features of Python and Django.

Add an install option for the documentation

tested locally by building successfully


clintonb (approved):

ulgens:
@dancergraham When possible, always try to rebase instead of merging in so branch histories can stay linear.

Issue #510, opened 2024-11-02
Feature request: Add template filters for switch_is_active, flag_is_active, sample_is_active


MatthewEthanTam:
In my usage of Django-waffle, I found that I cannot use the switch template tag within an if statement's condition in Django templates as some blocks of code require both the switch to be active and another context variable to be set.

Please find an example of this case below:
```
{% switch 'switch'%}
  {% if context_variable %}
    modified message
  {% else %}
    default message
  {% endif %}
{% else %}
  default
{%endswitch%}
```

Suggestion: I suggest we make template filters for `switch_is_active`, `flag_is_active`, and `sample_is_active`, this will allow the reduction of redundant code in this scenario.

an example of this suggestion is below:
```
{% if 'switch'|switch_is_active and context_variable%}
  modified message
{% else %}
  default message
{% endif %}
```

If this is a good suggestion, I can make a PR for it.

steynvl:
Would find use for this in my projects using `django-waffle`

ghost:
Hi! I'm just starting with open source and I'd love to give this a try.

ghost:
@clintonb @dancergraham bump

dancergraham:
Hello, I don't really use templates so I don't have an opinion on this and wouldn't be much help getting started on a PR - I would start by putting a question on the django forum to see whether anyone has a good shorthand for this using existing template features such as https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#with with statements (random idea - don't know whether it is relevant.

MatthewEthanTam:
Thank you for the suggestion @dancergraham, @whoiskrtk2 I have already forked the repo and have a PR feature branch ready, I am happy to walk the PR through with you 👍🏻

PR #532, opened 2024-11-18
Feature/adding template filter for switch flag sample usable for if statement conditions

MatthewEthanTam:
This PR is addressing #510, as the previous PR was out of sync

### DIFF
--- a/waffle/templatetags/waffle_tags.py
+++ b/waffle/templatetags/waffle_tags.py
@@ -1,7 +1,13 @@
+from typing import Optional
 from django import template
+from django.http import HttpRequest
 from django.template.base import VariableDoesNotExist
 
-from waffle import flag_is_active, sample_is_active, switch_is_active
+from waffle import (
+    flag_is_active as waffle_flag_is_active,
+    sample_is_active as waffle_sample_is_active,
+    switch_is_active as waffle_switch_is_active
+    )
 from waffle.views import _generate_waffle_js
 
 
@@ -60,17 +66,17 @@ class WaffleNode(template.Node):
 
 @register.tag
 def flag(parser, token):
-    return WaffleNode.handle_token(parser, token, 'flag', flag_is_active)
+    return WaffleNode.handle_token(parser, token, 'flag', waffle_flag_is_active)
 
 
 @register.tag
 def switch(parser, token):
-    return WaffleNode.handle_token(parser, token, 'switch', lambda request, name: switch_is_active(name))
+    return WaffleNode.handle_token(parser, token, 'switch', lambda request, name: waffle_switch_is_active(name))
 
 
 @register.tag
 def sample(parser, token):
-    return WaffleNode.handle_token(parser, token, 'sample', lambda request, name: sample_is_active(name))
+    return WaffleNode.handle_token(parser, token, 'sample', lambda request, name: waffle_sample_is_active(name))
 
 
 class InlineWaffleJSNode(template.Node):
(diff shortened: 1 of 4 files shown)


MatthewEthanTam:
@ulgens Please see new PR rebased to master and removed updates to .po files.

waffle/templatetags/waffle_tags.py
```diff
@@ -81,3 +87,32 @@ def render(self, context):
...
+@register.filter(name="switch_is_active")  # type: ignore[misc]
+def switch_is_active(switch_name: str) -> bool:
+    """
+    This template filter works like the switch tag, but you can use it within
+    if statement conditions in Django templates. Example: `{% if "switch_name"|switch_is_active %}`.
+    """
+    return waffle_switch_is_active(switch_name)
+
+
+@register.simple_tag  # type: ignore[misc]
```
dancergraham, line 101: question: should this have a `name` similar to `switch_is_active` and `sample_is_active`?

dancergraham, line 101: question: should this be a `register.filter` like the switch and sample above and below ? context: I don't use django templates so apologies if I'm missing something obvious

MatthewEthanTam: Good question @dancergraham , so since there are more than 2 arguments with differing types for this function, we cannot use a template filter for it, therefore, we use a simple tag and set the variable in the django template.

```
{% flag_is_active "flag" request True as is_flag_active %}
{% if is_flag_active %}
  flag_is_active on
{% else %}
  flag_is_active off
{% endif %}
```

MatthewEthanTam: Sadly it is not as clean of an implementation, and there are forums about this exact issue dating back to 2007 (https://code.djangoproject.com/ticket/1199)

ulgens:
@MatthewEthanTam I checked the PR and issue several times and I still don't understand the case here. I'll check it again when I have better availability. Sorry for the delay.

MatthewEthanTam:
@ulgens Thank you for your time and no worries about the delay, I am happy to contribute!

There are two cases. 

First, the case explained in the issue is where code is duplicated when a Switch and a context variable are needed to display text/ elements in an HTML block. The implementation will allow for cleaner HTML code and reduced lines of code.

Before implementation:
```
{% switch 'switch'%}
  {% if context_variable %}
    modified message
  {% else %}
    default message
  {% endif %}
{% else %}
  default message
{%endswitch%}
 ``` 
After implementation:
```
{% if 'switch'|switch_is_active and context_variable%}
  modified message
{% else %}
  default message
{% endif %}
```

Second, HTML linters/formatters such as DJLint do not recognize the `{% switch 'switch_name' %}` template tag, but do recognize `{% if switch_is_active|'switch_name' %}`, allowing for a better linting & formatting experience.

clintonb (commented):
Can you add documentation, please?

dancergraham:
@MatthewEthanTam do you need any help putting together the documentation?

MatthewEthanTam:
> @MatthewEthanTam do you need any help putting together the documentation?

Yes please is there a specific place I can document this implementation? @dancergraham

dancergraham:
> Yes please is there a specific place I can document this implementation? @dancergraham

I suggest putting it at the end of the templates page - A short explanation together with your "after" snippet from the discussion above would be good - let me know if you need some help. The ci will check that it builds OK for you or you can install sphinx and build it locally to test...

https://github.com/jazzband/django-waffle/blob/master/docs/usage/templates.rst

https://waffle.readthedocs.io/en/stable/usage/templates.html

MatthewEthanTam:
Thanks @dancergraham , I have added the documentation to the `templates.rst` file.

Issue #526, opened 2024-11-15
Improve api documentation


dancergraham:
The api is only partially documented in the online docs, for instance the flags page lists attributes (but capitalises them) and does not list methods - it would be good to document the public methods

https://waffle.readthedocs.io/en/stable/types/flag.html

[Switches ](https://waffle.readthedocs.io/en/stable/types/switch.html)lists neither attributes nor methods

[Samples](https://waffle.readthedocs.io/en/stable/types/sample.html) lists attributes (but capitalises them) and does not list methods

yanurag-dev:
@dancergraham I would like to pick this issue.

dancergraham:
Great, let me know if you need any help getting started

yanurag-dev:
@dancergraham Hi, I'm currently busy with a task and unable to contribute this time, so I will unassign myself.

PR #536, opened 2024-11-26
Test all pull requests for docs

dancergraham:
This workflow runs against all PRs to catch problems with docs. It catches any problems before merging to master. 

Related to #526

  update the CI workflow to include a documentation build stage.


clintonb (commented):
This should be done for pull requests that touch docs. Finding a failure after merging to master is too late. Can you add a workflow that runs against PRs?

dancergraham:
> This should be done for pull requests that touch docs. Finding a failure after merging to master is too late. Can you add a workflow that runs against PRs?

Hello, this workflow runs against both PRs and on merge - see here 
https://github.com/jazzband/django-waffle/actions/runs/12074017053/job/33671235325?pr=536

dancergraham:
@clintonb could you look again please - I think this PR already does what you are asking for:

* Without this PR doc building can fail silently on readthedocs.io - the only visible sign will be the new status badge on the readme
* With this PR the CI will fail before merging if the docs do not build correctly

dancergraham:
Hi @clintonb I still think this PR does exactly what you are asking for - can you please rephrase if I'm missing something?

clintonb (approved):

PR #539, opened 2024-12-03
Docs improve api docs

### LINKED ISSUE #526: Improve api documentation

dancergraham:
Add missing attributes and methods


docs/conf.py
```diff
@@ -119,7 +119,7 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+# html_static_path = ['_static']
```
clintonb, line 122: Why was this removed?

dancergraham: Hi, thanks for looking at this - this line raises a warning during build because there is nothing in there - I just commented it to make it easy to find again later if we want to add any static assets

dancergraham: You can see the warning on line 7 of the `Build docs` phase of my other PR, which adds tests for pull requests that touch docs

https://github.com/jazzband/django-waffle/actions/runs/12378362580/job/34550046030

clintonb (approved):

PR #544, opened 2024-12-30
fix: building sphinx rtd theme

dancergraham:
Better follow the rtd config instructions from 
https://docs.readthedocs.io/en/stable/config-file/v2.html#packages
This has the advantage of not duplicating the requirements specification


dancergraham:
Sorry @clintonb I think I just understood your comment on the docs ci job! Hopefully this will fix it...

clintonb (approved):

PR #547, opened 2025-01-16
Add code coverage calculation

### LINKED ISSUE #541: Needs code coverage calculation
relates to #501 

It would be great to have a code coverage calculation in the repo. My first thoughts on this:

* add coverage.py as a test dependency?
* configure coverage to check appropriate files e.g.:
  * exclude docs or report them separately?
  * exclude test_app or report it separately?
  * Any other standard config or exclusions for django projects?
* add a coverage section to the tox configuration, probably just using the latest supported versions of python and django? 
* ensure this runs properly in ci - add necessary secrets or env vars
* [if-minor]add a badge to the main project readme either linked to the github workflow or linked to codecov
* [if-minor]ensure it runs properly in codecov - needs config like this ?   https://github.com/jazzband/django-axes/blob/master/codecov.yml

dancergraham:
Add code coverage
Add upload to codecov - working here https://app.codecov.io/gh/jazzband/django-waffle/commits
Add badge to readme

fixes #541 , relates to #501

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -147,7 +147,7 @@ jobs:
           source .venv/bin/activate
           export DJANGO_SETTINGS_MODULE="test_settings"
           export PYTHONPATH=.".:$PYTHONPATH"
-          python -m coverage run --source=waffle django-admin test waffle
+          python -m coverage run --source=waffle `which django-admin` test waffle
           coverage report -m
       - name: Upload coverage to Codecov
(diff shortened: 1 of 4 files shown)


dancergraham:
Uses a virtual environment and calls coverage via python because:

* coverage seems to want a python command to run, not a shell command
* finding the coverage path and then getting coverage to find the django-admin path was a bit tricky. I couldn't get it to work with just `coverage run...`

clintonb (approved):

PR #549, opened 2025-02-24
Add ruff rules A,B,C4,EM,ISC,PERF,PGH,RET,RUF,S,SIM,SLF

cclauss:
https://docs.astral.sh/ruff/rules

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,13 +71,17 @@ target-version = "py38"
 
 [tool.ruff.lint]
 select = [
+  "A",      # flake8-builtins
   "AIR",    # Airflow
   "ASYNC",  # flake8-async
+  "B",      # flake8-bugbear
   "BLE",    # flake8-blind-except
+  "C4",     # flake8-comprehensions
   "C90",    # McCabe cyclomatic complexity
   "DJ",     # flake8-django
   "DTZ",    # flake8-datetimez
   "E",      # pycodestyle errors
+  "EM",     # flake8-errmsg
   "F",      # Pyflakes
   "FIX",    # flake8-fixme
   "FLY",    # flynt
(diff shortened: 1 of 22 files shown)


pyproject.toml
```diff
@@ -71,26 +71,38 @@ target-version = "py38"
...
   "E",      # pycodestyle errors
+  "EM",     # flake8-errmsg
   "F",      # Pyflakes
   "FIX",    # flake8-fixme
   "FLY",    # flynt
   "G",      # flake8-logging-format
   "ICN",    # flake8-import-conventions
   "INP",    # flake8-no-pep420
   "INT",    # flake8-gettext
+  "ISC",    # flake8-implicit-str-concat
```
dancergraham, line 92: The above lints look useful to me in that the changes made to make the lints pass mostly look like useful improvements. The possible exception to my mind is defining a `msg` variable before raising the exception - I am ambivalent to that one.

pyproject.toml
```diff
@@ -71,26 +71,38 @@ target-version = "py38"
...
   "NPY",    # NumPy-specific rules
   "PD",     # pandas-vet
+  "PERF",   # Perflint
```
dancergraham, line 95: perflint describes itself as "This project is an early beta. It will likely raise many false-positives in your code" - you have previously said that you think any and all passing lints should be included - more is necessarily better. 
I am worried that new linting raises the bar for new contributors, making it harder to get started - linting has advantages and disadvantages.
If any lints should be excluded then I think this one should, given the warning from its creator.

pyproject.toml
cclauss: % `ruff rule EM101`
> ___Why is this bad?___
By using a string literal, the error message will be duplicated in the traceback, which can make the traceback less readable.

If you would like, we can revert the EM transformations.

% `ruff check --select=EM--statistics`
```
10	EM101  	[*] raw-string-in-exception
 6	EM102  	[*] f-string-in-exception
[*] fixable with `ruff check --fix`
```

pyproject.toml
cclauss: This pull request did not make any PERF-related changes (other than enabling PERF).

When I read the ___Why is this bad?___ section of [these rule](https://docs.astral.sh/ruff/rules/#perflint-perf) I see quantified performance gains.

Here we are not running `perflint`.  Instead, we are running `ruff` which uses the original project as inspiration but rewrites each rule that the project team deems valuable in Rust.  Often useless or conflicting rules are dropped.  There is active discussion in the `ruff` repo about which rules to adopt and which to drop.

If you would like, we can disable the PERF rules.

PR #554, opened 2025-04-23
Extracted percent evaluation to separate function

longchiwen:
This update will extract current percent and rollout functionality to a separate function, so it can be extended on a custom feature flag models. No changes in functionality was introduced. It follows same convention as `_is_active_for_user` and similar functions.


clintonb (approved):

longchiwen:
Would appreciate the merge when you have the time 🙏

PR #561, opened 2025-06-05
Fixing the issue of everyone setting not being taken into account by flag.is_actvive_for_user method.

imohib:
Currently, if I read the documents correctly, it reads that `everyone` setting/config. for a waffle flag should override all the other settings when set to `True` (i.e. 'Yes') or `False` (i.e. 'No'), but current version of the code contradicts this.

I have added the testcase `WaffleFlagEveryoneSettingTests` to confirm that is the case. 
Solution was to update the method `is_active_for_user` to also consider the case where `everyone` is set to `False`.

Raising this PR to fix the issue. Any questions, please let me know.

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -230,8 +230,9 @@ class AbstractBaseFlag(BaseModel):
         return flush_keys
 
     def is_active_for_user(self, user: AbstractBaseUser) -> bool | None:
-        if self.everyone:
-            return True
+        # According to the docs, everyone should override all the other settings
+        if self.everyone is not None:
+            return self.everyone
 
         if self.authenticated and user.is_authenticated:
             return True
--- a/waffle/tests/test_waffle.py
+++ b/waffle/tests/test_waffle.py
@@ -803,3 +803,39 @@ class FunctionTests(TestCase):
         response = process_request(request, views.flag_in_view_readonly)
         assert 'dwf_myflag' not in response.cookies
         self.assertEqual(b'off', response.content)
+
+
+class WaffleFlagEveryoneSettingTests(TestCase):
+    databases = DATABASES
+
+    def test_is_active_for_user_respects_everyone_on(self):
+        """
+        Test flag.is_active_for_user returns truthy value when everyone is set to True overriding all other settings.
+        """
+        flag = waffle.get_waffle_flag_model().objects.create(
+            name="feature_flag_name",
+            staff=False,
+            everyone=True,
+        )
+        staff_user = get_user_model()(
+            id=999,
+            username="foo",
+            is_staff=True,
+        )
+        self.assertTrue(flag.is_active_for_user(staff_user))
+
+    def test_is_active_for_user_respects_everyone_off(self):
+        """
+        Test flag.is_active_for_user returns falsy value when everyone is set to False overriding all other settings.
+        """
+        flag = waffle.get_waffle_flag_model().objects.create(
+            name="feature_flag_name",
+            staff=True,
+            everyone=False,
+        )
+        staff_user = get_user_model()(
+            id=999,
+            username="foo",
+            is_staff=True,
+        )
+        self.assertFalse(flag.is_active_for_user(staff_user))


imohib:
@clintonb, any chance of changes in this PR getting rolled into version `4.2.1`? 

image

dancergraham:
I don't think this can be in a minor update as it changes the current behaviour (albeit to match a literal reading of the docs)

An alternative without changing current behaviour would be to update the docs to say "Overiding all other criteria if true".

I'm not sure what the expected/desired behaviour should be if you set everyone to True, then set everyone to False, then give the flag to a subset of users ?

dancergraham:
hmm looks like a duplicate of https://github.com/jazzband/django-waffle/pull/498

clintonb:
Yes, this is a breaking change that must be released as a new major version.

imohib:
> I don't think this can be in a minor update as it changes the current behaviour (albeit to match a literal reading of the docs)

I'd assume docs would be the first place you'd read to know what's available to people as application developers. Very rarely, I'd expect people to jump straight into the source code for every package or library they make use of or exploring to make use of.

> An alternative without changing current behaviour would be to update the docs to say "Overiding all other criteria if true".

That to me feels like updating the acceptance criteria retrospectively to match the code. As in - it's a feature, not a bug&trade; 😄 

> I'm not sure what the expected/desired behaviour should be if you set everyone to True, then set everyone to False, then give the flag to a subset of users ?

To me at least, everyone implies a `set` containing every single member conceivable. A `superset`. An exhaustive list of members. If we set it to `True` for everyone, then it should be `True` for `100%` of the users, and `False` otherwise (i.e. `0%` of the users). 

In the case of 'unknown' (i.e. `None`), other attributes/settings/parameters should determine whether a flag should evaluate to `True` or `False`.

@dancergraham, for breaking change, I'd expect at least one of the tests to fail. Do we have a unittest to ensure it fails if/when the changeset proposed gets merged to make it explicit that this is indeed a feature and not a bug?

imohib:
> Yes, this is a breaking change that must be released as a new major version.

@clintonb, 
😢. Please, consider at least adding a unittest so that indeed the changeset proposed here will make it fail.
Perhaps updated docs to go with it with a caveat to note something along the line that reads - this was a typo, expect this to change in version X.Y.Z ... 

Also, is it safe to assume that a new major version will drop support for Python 3.8 making current version `4.2.0` the last version to support it?

I am happy to raise another PR to action just this unless we already have PR(s) that do just that.

PS - This is really a great package. Donation on the way 😉. Keep up the good work!

clintonb:
@imohib please add a test. It's your change. I'll drop EOL Python and Django versions in a separate PR. Will merge and release sometime tomorrow.

imohib:
> @imohib please add a test. It's your change. I'll drop EOL Python and Django versions in a separate PR. Will merge and release sometime tomorrow.

@clintonb (+ @dancergraham), done. 👉 #562

clintonb:
@imohib the tests should be part of _this_ PR since this one changes the behavior.

imohib:
> @imohib the tests should be part of _this_ PR since this one changes the behavior.

@clintonb, ok, there seems to be some confusion here. there are two PRs as it stands and they both come with unittests.  

1. [One PR](https://github.com/jazzband/django-waffle/pull/562) is proposing that we update the docs (as suggested by @dancergraham in this comment thread [here](https://github.com/jazzband/django-waffle/pull/561#issuecomment-2949428653)) and comes with [the testcase](https://github.com/jazzband/django-waffle/pull/562/files#diff-9d717bc52b672de0430e050db9372eaa0e715a8d21bf90bbb0ff22326e27eb50R808-R912) that makes it clear what the behaviour should be, so if a breaking change is introduced, CI test run would fail.
2. Then there is [this one](https://github.com/jazzband/django-waffle/pull/561) that is proposing that we update the code to change the behaviour, which is a breaking change and needs to go out in a major release and not as a patch and comes with [a testcase to confirm the change](https://github.com/jazzband/django-waffle/pull/561/files#diff-9d717bc52b672de0430e050db9372eaa0e715a8d21bf90bbb0ff22326e27eb50R808-R841).

Q.1: Do we want to release #562 first to the current version so that new/existing users are aware that docs have been updated and what the expected behaviour should be? 

Q.2: Are we saying, no need to release #562, just release #561 (this one) as its major new release? It does come with [the testcase](https://github.com/jazzband/django-waffle/pull/561/files#diff-9d717bc52b672de0430e050db9372eaa0e715a8d21bf90bbb0ff22326e27eb50R808-R841).

clintonb:
Skip #562. We’ll release one new major version with #561 and highlight the
fact that the implementation now matches the documentation in the release
notes.

clintonb (approved):

PR #563, opened 2025-06-08
Dropped support for Python 3.8 and Django 5.0

clintonb:
Both Python 3.8 and Django 5.0 have reached EOL, so are no longer supported by this package.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,13 +18,11 @@ classifiers = [
     "Operating System :: OS Independent",
     "Framework :: Django",
     "Framework :: Django :: 4.2",
-    "Framework :: Django :: 5.0",
     "Framework :: Django :: 5.1",
     "Framework :: Django :: 5.2",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3 :: Only",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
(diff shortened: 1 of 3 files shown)


clintonb:
Well, crap! @jezdez has yet to respond to either https://github.com/jazzband/django-waffle/issues/500 or https://github.com/jazzband/help/issues/196, so I cannot remove the test requirement for Python 3.8. I will revisit migrating this project to https://github.com/django-commons tomorrow/later this week.

pyproject.toml
```diff
@@ -68,7 +66,7 @@ ignore_missing_imports = true
 
 [tool.ruff]
 line-length = 120
-target-version = "py38"
+target-version = "py313"
```
dancergraham: should be py39 to avoid ruff suggesting incompatible upgrades

pyproject.toml
```diff
@@ -55,7 +53,7 @@ namespaces = false
 waffle = ["py.typed"]
 
 [tool.mypy]
-python_version = "3.8"
+python_version = "3.13"
```
dancergraham: should be 3.9 to avoid mypy failing on older syntax

dancergraham (changes requested):

dancergraham:
@clintonb I think the ci failures are coming from specifying too high a python version for linting - fixing this should clear most or all of the errors.

hugovk:
> Well, crap! @jezdez has yet to respond to either #500 or [jazzband/help#196](https://github.com/jazzband/help/issues/196), so I cannot remove the test requirement for Python 3.8.

In the meantime, you can work around it by having the 3.8 job be a no-op that doesn't test anything.

jezdez:
Looking later today

dancergraham (approved):

ulgens:
@clintonb We already have a ready PR for 3.8: https://github.com/django-waffle/django-waffle/pull/529/ Is something wrong with that PR, I can update it if necessary.

clintonb:
See https://github.com/django-waffle/django-waffle/pull/529

PR #574, opened 2025-12-02
Fix: is_active_for_user should be False when everyone is false even if users or groups are added

lrivkin:
Make the "everyone" setting override all other settings including if a user or group is specifically added to the flag through .user or .groups.

Relates to https://github.com/django-waffle/django-waffle/pull/561 but fixes the issue if "everyone" is used in addition to adding users or groups to the `.users` or `.groups` set of a user-based flag. 

I'm not sure if this counts as a "breaking" change – it seems like it gets to completely fixing the behavior that was fixed in 561 but for flags that target users and groups (not just the staff, superuser, etc. settings)

### DIFF
--- a/docs/types/flag.rst
+++ b/docs/types/flag.rst
@@ -55,18 +55,18 @@ Flags can be administered through the Django `admin site`_ or the
     A percentage of users for whom the flag will be active, if no other
     criteria applies to them.
 :superusers:
-    Is this flag always active for superusers?
+    Is this flag active for superusers?
 :staff:
-    Is this flag always active for staff?
+    Is this flag active for staff?
 :authenticated:
-    Is this flag always active for authenticated users?
+    Is this flag active for authenticated users?
 :languages:
     Is the ``LANGUAGE_CODE`` of the request in this list?
     (Comma-separated values.)
 :groups:
-    A list of group IDs for which this flag will always be active.
+    A list of group IDs for which this flag is active.
 :users:
-    A list of user IDs for which this flag will always be active.
+    A list of user IDs for which this flag is active.
 :rollout:
     Activate Rollout mode? :ref:`See below <types-flag-rollout>`.
 :note:
--- a/waffle/tests/test_waffle.py
+++ b/waffle/tests/test_waffle.py
@@ -291,11 +291,22 @@ class WaffleTests(TestCase):
         self.assertEqual(b'off', response.content)
         assert 'dwf_myflag' not in response.cookies
 
-    def test_percent(self):
-        """If you have no cookie, you get a cookie!"""
+    @mock.patch.object(random, 'uniform')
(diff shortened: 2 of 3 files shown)


waffle/tests/test_waffle.py
```diff
@@ -291,11 +291,22 @@ def test_everyone_off(self):
...
         request = get()
+        uniform.return_value = '10'  
         response = process_request(request, views.flag_in_view)
+        self.assertEqual(b'on', response.content)
+        assert 'dwf_myflag' in response.cookies
+
+        # > 50. Flag is False.
+        request = get() # Fresh request
+        uniform.return_value = '80'
+        response = process_request(request, views.flag_in_view)
```
lrivkin, line 308: This explicitly tests the "off" scenario for % based requests sets a cookie which became uncovered when the "everyone=False" behavior changed. 

Happy to add this as a different test or move around somewhere else if this test should stay as is!

waffle/tests/test_waffle.py
```diff
@@ -839,3 +850,12 @@ def test_is_active_for_user_respects_everyone_off(self):
             is_staff=True,
         )
         self.assertFalse(flag.is_active_for_user(staff_user))
+
+    def test_is_active_for_user_respects_everyone_off_for_targeted_users(self):
```
lrivkin, line 854: This was my regression test here - this failed with the current version and then passed with the change applied to `is_active_for_user`

docs/types/flag.rst
```diff
@@ -55,18 +55,18 @@ Flags can be administered through the Django `admin site`_ or the
     A percentage of users for whom the flag will be active, if no other
     criteria applies to them.
 :superusers:
-    Is this flag always active for superusers?
+    Is this flag active for superusers?
```
lrivkin, line 58: The docs already say "everyone overrides all other criteria" - so removing the word "always" here I think makes these docs a tiny bit more clear.

```
everyone: False
superusers: True
```
will mean even superusers will see `False` for `flag_is_active` so the flags are NOT "always active" for superusers

clintonb (approved):
This seems like a good fix. The docs explicitly mention that this is a three-phase control—on, off, unknown:

> Globally set the Flag, overriding all other criteria. Leave as Unknown to use other criteria.

https://waffle.readthedocs.io/en/stable/types/flag.html#how-flags-work

This is introducing a breaking change, so will be released as part of a new major version.

lrivkin:
Sounds good! I wasn’t sure if it would count as a “bug fix” of the breaking change in 5 for the similar issue or its own breaking change. 

Is there anything else I need to keep an eye on for this branch or will it just get merged when the next major release gets cut?

gregjkal:
We were bit by this when using `@override_flag` in tests, since that utility just sets the value for `everyone`, leaving the rest of the values in place.

PR #582, opened 2026-02-23
feat: Cache operations fall back to db when cache is inaccessible

### LINKED ISSUE #399: Catch cache errors and raise errors where necessary
Waffle currently tries to retrieve a setting from the cache, then goes to the DB if there is a cache miss. This doesn't account for issues accessing the cache. If there is a cache issue (e.g. a `TimeoutError`), this error is thrown to the caller. In prod, we've encountered several of these issues for a small percentage of our traffic. 

I'd suggest making the following changes:
- By default, catch exceptions thrown by the cache and go directly to the database.
- Add a setting that allows the dev to configure the number of retries. 
- Allow the dev to configure the number of retries when calling each function (e.g. `switch_is_active`)

scitech:
Attempts to resolve #399 

Previously when a cache is down, waffle would propagate e.g. a `RedisConnectionError`. For our application, this surfaced as an outage for end users.

With this PR, when a cache is down, waffle will handle the e.g. `RedisConnectionError` and look in the database for a flag value instead.

Adds two new settings: 
* CACHE_FAILOVER (default True) - whether to failover to the db in the event of an exception trying to perform a cache operation
* CACHE_RETRIES (default 0) - number of retry attempts after a failed cache operation, before failing over to the db


elialbert:
CACHE_FAILOVER default True would be a big change for existing users - I would think defaulting False and making this opt-in would be a smoother transition for this change to be adopted.

hgbao:
I second this. However, agree with @elialbert , default to be `False` would be better for backward compatibility. Hope this is soon to be released! 🙏

oliverhaas:
Just passing by here, but I my initial feeling is  not to include any cache retry logic in django-waffle. 
Cache retries might fit better in a custom cache backend.

PR #587, opened 2026-04-06
fix: coverage exclude tests

oliverhaas:
Not sure if I'm just confused, but while working on #586 I noticed the coverage is also counted on the tests themselves, which does not make sense to me. So small suggestion for adjusting pyproject.toml...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,10 @@ namespaces = false
 [tool.setuptools.package-data]
 waffle = ["py.typed"]
 
+[tool.coverage.run]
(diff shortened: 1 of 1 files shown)


dancergraham:
I don't have a strong opinion on this but the creator of the coverage library does so I'm leaning towards keeping the current behaviour...  he https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage

oliverhaas:
Makes sense. Didn't know about the whole apparently existing debate and never questioned my view. Then nevermind!

PR #590, opened 2026-06-01
Prefetch flag users/groups to prevent N+1 queries

lightswitch05:
# Why

When `get_all()` loads flags from the database, it previously fetched each flag without prefetching the `users` and `groups` M2M relations. Any subsequent access to those relations — such as checking whether a flag is active for a given user or group — fired two additional queries per flag, producing an N+1 pattern that scaled with the number of flags.

# What

- Extract `_get_all_queryset` from `BaseModel.get_all_from_db` so subclasses can augment the queryset before it is materialised.
- Override `_get_all_queryset` on `AbstractUserFlag` to add `prefetch_related('users', 'groups')`, collapsing the per-flag M2M queries into two bulk queries regardless of flag count.
- Override `get_all` on `AbstractUserFlag` to batch-fetch the user and group waffle cache keys for all flags in a single `cache.get_many` call, and stamp each flag instance with `_prefetched_user_ids` / `_prefetched_group_ids` so that `_get_user_ids()` and `_get_group_ids()` return immediately on subsequent calls without touching the cache again.
- Merge the duplicated `_get_user_ids` / `_get_group_ids` cache-miss bodies into a shared `_get_relation_ids` helper.

# Testing

Tests assert that after `get_all_from_db`, accessing `.users` and `.groups` on flag instances requires zero additional queries. Further tests confirm that `_get_user_ids()` and `_get_group_ids()` also require zero queries when called on prefetched flags, that `_prefetch_user_group_ids` issues exactly one `cache.get_many` call regardless of flag count, and that the `CACHE_EMPTY` sentinel is handled correctly.

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -90,11 +90,15 @@ class BaseModel(models.Model):
         return objs
 
     @classmethod
-    def get_all_from_db(cls: type[_BaseModelType]) -> list[_BaseModelType]:
+    def _get_all_queryset(cls: type[_BaseModelType]) -> models.QuerySet[_BaseModelType]:
         objects = cls.objects
         if get_setting('READ_FROM_WRITE_DB'):
             objects = objects.using(router.db_for_write(cls))
-        return list(objects.all())
+        return objects.all()
+
+    @classmethod
+    def get_all_from_db(cls: type[_BaseModelType]) -> list[_BaseModelType]:
+        return list(cls._get_all_queryset())
 
     def flush(self) -> None:
         cache = get_cache()
@@ -364,6 +368,10 @@ class AbstractUserFlag(AbstractBaseFlag):
             keyfmt(get_setting('FLAG_GROUPS_CACHE_KEY'), self.name),
         ])
         return flush_keys
+
+    @classmethod
+    def get_all_from_db(cls: type[_BaseModelType]) -> list[_BaseModelType]:
+        return list(cls._get_all_queryset().prefetch_related('users', 'groups'))
 
     def _get_user_ids(self) -> set[Any]:
         cache = get_cache()
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -373,39 +373,28 @@ class AbstractUserFlag(AbstractBaseFlag):
     def get_all_from_db(cls: type[_BaseModelType]) -> list[_BaseModelType]:
         return list(cls._get_all_queryset().prefetch_related('users', 'groups'))
 
-    def _get_user_ids(self) -> set[Any]:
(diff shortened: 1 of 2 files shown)


waffle/models.py
```diff
@@ -365,39 +369,32 @@ def get_flush_keys(self, flush_keys: list[str] | None = None) -> list[str]:
...
         cache = get_cache()
-        cache_key = keyfmt(get_setting('FLAG_USERS_CACHE_KEY'), self.name)
+        cache_key = keyfmt(get_setting(cache_key_setting), self.name)
         cached = cache.get(cache_key)
         if cached == CACHE_EMPTY:
             return set()
         if cached:
             return cached
 
-        user_ids = set(self.users.all().values_list('pk', flat=True))
```
lightswitch05, line 409: Since users and groups are now prefetched, calling `values_list` would trigger an unnecessary DB query. I made the same change below in `_get_group_ids` - but then further refactored the logic to be reused into `_get_relation_ids`. I'm happy to revert this aspect of the PR if its overstepping.

waffle/models.py
```diff
@@ -365,39 +369,32 @@ def get_flush_keys(self, flush_keys: list[str] | None = None) -> list[str]:
         ])
         return flush_keys
 
-    def _get_user_ids(self) -> set[Any]:
+    @classmethod
+    def get_all_from_db(cls: type[_BaseModelType]) -> list[_BaseModelType]:
+        return list(cls._get_all_queryset().prefetch_related('users', 'groups'))
```
lightswitch05: Prefetch `users` and `groups`

waffle/models.py
```diff
@@ -90,11 +90,15 @@ def get_all(cls: type[_BaseModelType]) -> list[_BaseModelType]:
         return objs
 
     @classmethod
-    def get_all_from_db(cls: type[_BaseModelType]) -> list[_BaseModelType]:
+    def _get_all_queryset(cls: type[_BaseModelType]) -> models.QuerySet[_BaseModelType]:
```
lightswitch05, line 93: Expose this query set logic in `BaseModel` so that `get_all_from_db` can easily be overridden without needing to reimplement `READ_FROM_WRITE_DB` feature

lightswitch05:
Thanks for merging https://github.com/django-waffle/django-waffle/pull/591! Since that is now merged, this PR had some conflicts that I resolved - but also there were more opportunities to make use of the better cache logic. I've updated the PR accordingly. There are more changes in here than strictly necessary, I'm happy to scale back any unwanted changes or make any other requested changes. I think there is still some N+1 logic with `cache.add` that should be replaced with `cache.set_many` - but I need to spend more time on it.

PR #591, opened 2026-06-01
Batch-fetch user/group cache keys in get_all() to fix N+1

lightswitch05:
## Why

`waffle_json` (used by the `GET /feature_flags/waffle_status` endpoint) calls `get_all()` to load all flags, then evaluates each flag's user and group membership for the requesting user. Each of those evaluations triggered an individual `cache.get()` call — one for user IDs and one for group IDs per flag. On a deployment with many flags this produced an N+1 cache query pattern that was caught in Datadog.

## What

Override `get_all()` on `AbstractUserFlag` to batch-fetch all user and group cache keys in a single `cache.get_many()` call immediately after the flags are loaded. The results are stored on two new instance attributes (`_prefetched_user_ids`, `_prefetched_group_ids`). `_get_user_ids()` and `_get_group_ids()` short-circuit to return those values when present, skipping the per-flag cache lookup entirely.

Flags whose keys are absent from cache leave the attributes as `None`, so the existing individual-fetch fallback path is still reached (e.g. when a flag's cache entry hasn't been warmed yet). The `CACHE_EMPTY` sentinel is handled the same way it is in the existing code — it maps to an empty set.

## Testing

Six new tests cover the prefetch behaviour:

- Correct user and group IDs are returned after `get_all()` without any additional `cache.get()` calls
- The prefetch issues exactly one `cache.get_many()` call regardless of the number of flags, with the correct number of keys
- Setting `_prefetched_*` directly on a flag instance is respected by the getters (no cache hit)
- Flags with absent cache keys are left with `None` attributes (fallback path preserved)
- A `CACHE_EMPTY` sentinel in the cache yields an empty set on the attribute
- An empty flags list results in zero `cache.get_many()` calls

### DIFF
--- a/waffle/models.py
+++ b/waffle/models.py
@@ -352,6 +352,9 @@ class AbstractUserFlag(AbstractBaseFlag):
         verbose_name=_('Users'),
     )
 
+    _prefetched_user_ids: set | None = None
+    _prefetched_group_ids: set | None = None
+
     class Meta(AbstractBaseFlag.Meta):
         abstract = True
         verbose_name = _('Flag')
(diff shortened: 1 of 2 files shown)


lightswitch05:
Thinking more about this... I _do_ think this PR provides good value, but I'm not sure it goes far enough. On cache miss, it's going to have the same N+1 issue with cache set. I might put together another PR to be more holistic. Any feedback on your appetite to address these performance issues would be appreciated.

dancergraham:
Hello, yes avoiding an N+1 query sounds useful... I don't personally use that endpoint but using good Django patterns to avoid this alert would be good.

clintonb (approved):

lightswitch05:
This was merged a couple months ago. Would it be possible to get it included in an official release?

